PR #54, opened 2026-04-07
Add Telegram browser login for user credential extraction

joshalbrecht:
## Summary

- Adds `TelegramUserCredentials` type that stores MTProto `auth_key` data extracted from `web.telegram.org` localStorage after browser-based login
- `latchkey auth browser telegram` now opens a browser, waits for the user to log in, then extracts the auth_key and persists it in the encrypted credential store as `TelegramUserCredentials`
- Adds `latchkey auth get <service_name>` command that outputs stored credentials as JSON, enabling programmatic access from external scripts
- The existing `latchkey auth set-nocurl telegram <bot-token>` flow for bot tokens is unchanged

## Changes

* New `TelegramUserCredentials` class storing `dcId`, `authKeyHex`, `userId`, `firstName`
* New `TelegramBrowserSession` that overrides `login()` with a manual "press Enter after login" flow (Telegram Web A's SPA makes response-based auto-detection unreliable)
* `TelegramUserCredentialsSchema` registered in `apiCredentialsSerialization.ts`
* `Telegram.checkApiCredentials()` returns `Unknown` for user credentials (MTProto can't be checked via curl)
* New `auth get` CLI command for programmatic credential retrieval
* Registry test updated: uses AWS instead of Telegram as example of service without `getSession`

## Test plan

- [x] `npm run build` passes
- [x] `npm run lint` passes
- [x] `npm run format:check` passes
- [x] `npm test` passes (265 tests pass; 2 timeout-only failures on lint/typecheck test wrappers that pass when run directly)
- [x] Manual test: ran `latchkey auth browser telegram`, logged in, verified credentials extracted and persisted
- [x] Manual test: used extracted credentials with Telethon to connect and send messages to BotFather
- [x] Manual test: `latchkey auth get telegram` outputs stored credentials as JSON

Generated with [Claude Code](https://claude.com/claude-code)


hynek-urban:
Thanks, Josh :) I appreciate this! Still, I think most of this functionality is probably out of scope for Latchkey:

- `latchkey auth get` is missing intentionally. Latchkey is rather specialized - it injects API credentials to curl-encoded HTTP requests. In other words, it is not a generic credentials management tool. It's meant to be used by agents and we don't want those to be able to retrieve arbitrary credentials in plaintext. (Ideally, there should exist better guarantees than just a missing method in the interface but even so, the direction we'd like to go is the opposite of `auth get` :) )
- The newly introduced `auth browser telegram` runs an interactive session in the terminal (waiting for keypresses) which makes it different from all other `auth browser` invocations and unsuitable e.g. as an `llm` tool.
- The implementation returns credentials that can only be used with the MTProto protocol, not HTTP, making it out of scope for a "HTTP credentials injection" tool.

PR #63, opened 2026-05-04
Add notion-mcp service for OAuth 2.0 with PKCE

boweiliu:
## Why

The existing `notion` service creates internal integration tokens (`ntn_...`) via browser automation. These tokens work with `api.notion.com` but are very limited in page access -- can't search, have to be manually added to files etc.

Notion's MCP server at `mcp.notion.com` exposes a much richer toolset (search across workspace + connected sources, page fetching, database operations, comments, users) but requires OAuth 2.0 authentication. Internal integration tokens are rejected. Since these are fundamentally different auth flows and credential types, `notion-mcp` is a separate service.

## Implementation

Standard OAuth 2.0 authorization code flow with PKCE, following the existing Google OAuth patterns in the codebase. Written by an LLM.

Changes:
- **`oauthUtils.ts`**: Added PKCE helpers (`generateCodeVerifier`, `generateCodeChallenge`), optional `codeVerifier` param to `exchangeCodeForTokens()`, and made `client_secret` conditional in both token exchange and refresh (for public clients)
- **`services/notion-mcp.ts`**: New service — dynamic client registration via HTTP POST (no browser automation needed), PKCE OAuth flow, token refresh
- **`services/index.ts`** + **`serviceRegistry.ts`**: Wired into registry

The OAuth changes to `oauthUtils.ts` are backward-compatible and will benefit any future service that needs PKCE.

## Usage

MCP is a session-based protocol, so curl usage requires three steps:
1. `latchkey auth browser notion-mcp` — authenticate once via OAuth
2. `latchkey curl -X POST -H "Accept: application/json, text/event-stream" -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"latchkey","version":"1.0"}},"id":1}' -D - https://mcp.notion.com/mcp` — initialize a session and grab the `Mcp-Session-Id` from the response headers
3. `latchkey curl -X POST -H "Accept: application/json, text/event-stream" -H "Content-Type: application/json" -H "Mcp-Session-Id: <session-id>" -d '{"jsonrpc":"2.0","method":"tools/call","params":{"name":"notion-search","arguments":{"query":"my search","query_type":"internal"}},"id":2}' https://mcp.notion.com/mcp` — make tool calls using that session ID

## Test plan

- [ ] Run `notion-mcp` auth flow end-to-end, verify token stored
- [ ] Verify existing `notion` service still works unchanged
- [ ] Verify token refresh works for `notion-mcp`
- [ ] Confirm other OAuth services (Google) unaffected by `oauthUtils` changes


src/services/notion-mcp.ts
```diff
@@ -0,0 +1,237 @@
...
+        tokens.access_token,
+        tokens.refresh_token ?? apiCredentials.refreshToken,
+        accessTokenExpiresAt,
+        apiCredentials.refreshTokenExpiresAt
+      )
+    );
+  }
+}
+
+export const NOTION_MCP = new NotionMcp();
```
boweiliu, line 246: bowei: i read this code manually, doesn't seem like the greatest code  in the world but works e2e. doesn't seem to handle any timeouts or error recovery in general

src/oauthUtils.ts
```diff
@@ -211,9 +233,11 @@ export function refreshAccessToken(
   const body = new URLSearchParams({
     refresh_token: refreshToken,
     client_id: clientId,
-    client_secret: clientSecret,
     grant_type: 'refresh_token',
   });
+  if (clientSecret) {
+    body.set('client_secret', clientSecret);
+  }
```
boweiliu, line 240: bowei: i read this code, slightly sus but also seems pretty fine/standard

hynek-urban:
@boweiliu Thanks for the contribution :) This is funny. I used to think that letting agents use public HTTP APIs directly to circumvent MCP was a good thing and now we're using MCP to circumvent the issues with the "normal" public HTTP API :)

A few questions:

- Did you test that google services still work after the change in oauthUtils.ts?
- Having implemented this, do you think there's a reason to keep the "normal" notion integration in place? As in, are there any use cases that can be served with the "normal" integration but not this one?
- Could you please expand the description in the NotionMcp's `info` string? It's a string that agents get to read while examining available services so by reading `info` of both the "old" `Notion` and of `NotionMCP`, they should be able to make a decision about which one to use.

Also, the `info` string usually contains pointers to a documentation that the agent can use to be able to actually use the API. This is maybe my only potential objection to this MR: is there a good documentation for the MCP server when you want to use it directly over http? (We generally avoid including private or undocumented APIs because agents have no way of knowing how exactly to use them.)

boweiliu:
Added basic unit tests: smoke tests for `generateCodeVerifier` / `generateCodeChallenge`, plus body-construction tests for `exchangeCodeForTokens` (with/without PKCE) and `refreshAccessToken` (confidential/public client) that spy on `runCaptured` and assert on the URL-encoded `-d` arg.
Scope is body concoction only — the spy throws a sentinel to abort before any response parsing.

boweiliu:
made this change 0439e80:

**Why:** mcp spec was returning 406 which was causing latchkey auth to mark valid tokens as `invalid`.
**Change:** add the `Accept: application/json, text/event-stream` header (and a spec-shaped `params` block) to `credentialCheckCurlArguments` in `src/services/notion-mcp.ts`.
**Why missed:** forgot to run `latchkey auth list` and actually look at the json yesterday.

boweiliu:
@hynek-urban Thanks!

1. **Google OAuth** — manually re-tested by a human, still works.
2. **Keep the old Notion integration?** Yes — the REST API exposes things `mcp.notion.com` doesn't: per-block CRUD, file uploads, webhook management, token introspect/revoke, page-trash, granular page-property reads. `Notion.info` now marks it as limited-functionality and points at `notion-mcp` as the default; we can revisit removing it once MCP closes those gaps. Full write-up with the gap evidence: [`pr63-decision-keep-old-notion.md`](https://github.com/imbue-ai/latchkey/blob/wip/pr63-docs/pr63-decision-keep-old-notion.md) (also see the other docs in that branch).
3. **`NotionMcp.info`** — expanded on this branch. I also tested with an agent to "set up notion" and it correctly defaulted to the mcp version.

hynek-urban:
@boweiliu Cool, thanks! I'll merge this soon to the next release together with some other work that's currently in progress.

PR #73, opened 2026-05-20
Add additional base API URL for Slack service

thad-imbue:
This is necessary to curl files from Slack.


hynek-urban:
@thad-imbue Thanks!

PR #74, opened 2026-05-21
Remove broken browser automation from notion service

qi-imbue:
## Summary
- Remove the Playwright-based `NotionServiceSession` from the `notion` service since the internal-integration login flow is broken. The service now has no `getSession`, so the rest of the codebase treats it as having no automated login.
- Update `notion`'s `info` string to direct agents to use `notion-mcp` unless valid credentials are already set for `notion` (credentials can still be set manually by the user).
- The `notion-mcp` service is untouched.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx vitest run` passes (515 tests, including service registry, lint, typecheck)
- [ ] Manually verify `latchkey auth set notion -H "Authorization: Bearer <token>"` still works and that `latchkey` no longer offers a browser login for `notion`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "latchkey",
-  "version": "2.11.2",
+  "version": "2.11.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "latchkey",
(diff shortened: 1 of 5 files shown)


hynek-urban:
@qi-imbue Thanks!

PR #81, opened 2026-06-03
playwrightLoader: adapt to playwright-core 1.60 restructure

weishi-imbue:
playwright-core 1.60 dropped the internal subpaths we were importing in playwrightLoader.ts:

- `./lib/server/registry/index` (Registry instance)
- `./lib/zipBundle` (extract function)

Both files were removed AND not added to `package.json#exports`. Imports throw `ERR_PACKAGE_PATH_NOT_EXPORTED`, which our caller wraps as `BrowserFeaturesUnavailableError` — "Browser-based features are not available in the standalone latchkey binary". Misleading: latchkey isn't standalone, just incompatible with the new playwright-core.

This PR loads `lib/coreBundle` (which IS in the exports map) once and pulls the same symbols from the new locations:
- `coreBundle.default.registry.registry` → Registry instance
- `coreBundle.default.utils.extractZip` → extract function

The returned shape is unchanged so `playwrightDownload.ts` keeps working without changes.

Empirically verified locally that the new `registry` exposes `findExecutable('chromium')` and returns a real Executable. Diagnosed against the bundled `playwright-core@1.60.0` shipping inside the minds.app, which is the version most users encounter.

## Dependency bump (required)

`lib/coreBundle` only exists in playwright-core ≥ 1.60.0; 1.58.x/1.59.x still ship the old `lib/zipBundle` + `lib/server/registry/index` layout. latchkey declared `playwright ^1.58.2` (lockfile pinned 1.58.2), so the new imports would throw `ERR_PACKAGE_PATH_NOT_EXPORTED` on the `npm install -g latchkey` path even though they resolve fine against the minds.app-bundled 1.60. This PR therefore bumps `playwright` and regenerates the lockfile so latchkey's own declared dependency matches the layout the code now requires.

The range is pinned with a tilde (`~1.60.0`), not a caret. We import playwright-core *private internals*, which carry no semver guarantee and were restructured in the 1.59→1.60 minor bump. A caret would let a fresh `npm install -g latchkey` resolve a future minor that moves them again, silently reintroducing this exact failure. Tilde allows only 1.60.x patches; minor playwright upgrades become deliberate, re-verified changes.

## Typing

`playwright-core` ships no `.d.ts` for these internal subpaths, so `src/playwright-core.d.ts` carries an ambient declaration for `lib/coreBundle` (`Executable`, `Registry`, and the `default` export shape). `playwrightLoader.ts` derives its return types from that declaration — no inline casts or `@ts-expect-error`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@imbue-ai/detent": "^1.4.0",
         "@napi-rs/keyring": "^1.2.0",
         "commander": "^12.0.0",
-        "playwright": "^1.60.0",
+        "playwright": "~1.60.0",
         "zod": "^3.22.0"
       },
       "bin": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@imbue-ai/detent": "^1.4.0",
         "@napi-rs/keyring": "^1.2.0",
         "commander": "^12.0.0",
-        "playwright": "^1.58.2",
+        "playwright": "^1.60.0",
         "zod": "^3.22.0"
       },
       "bin": {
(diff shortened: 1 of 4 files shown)


hynek-urban:
Looks good to me. Thanks, @weishi-imbue !

PR #90, opened 2026-06-18
Configure the consent screen for reused Google projects

pseay-imbue:
## Problem

`latchkey auth browser-prepare google-gmail` failed with a Playwright timeout:

```
locator.waitFor: Timeout 12000ms exceeded.
  - waiting for locator('svg[data-icon-name="arrowDropDownIcon"]').first() to be visible
    6 × locator resolved to hidden <svg ... data-icon-name="arrowDropDownIcon">
  at createOAuthClient (.../services/google/base.js)
```

The cryptic timeout was a symptom. The real state of the page was Google's banner:

> ⚠️ To create an OAuth client ID, you must first configure your consent screen

## Root cause

To respect Google's per-account project cap, `browser-prepare` **reuses** a per-service project (e.g. `Latchkey-gmail`) across runs. But the OAuth consent screen (branding) was only configured on the **new-project** path — the reuse path skipped it entirely.

A project left half-configured by an earlier crashed run therefore never got its consent screen set up. On the next run the project was reused, branding was skipped, and `createOAuthClient` hit Google's wall. Because the banner replaced the form, the application-type dropdown never rendered, so the dropdown-arrow locator matched a *hidden* decorative SVG and timed out.

This mirrors a gap the code already acknowledged: it re-runs `enableApi` on every run to self-heal half-configured leftovers, but did not do the same for branding.

## Fix

- **Always configure the consent screen before creating the OAuth client**, for reused projects too (moved `configureBranding` out of the new-project-only branch).
- **Made `configureBranding` idempotent** — it detects whether the screen is already set up (empty-state "Get started" CTA vs. the populated branding form) and returns `null` (no-op) when it is, so running it on every run is safe. A reused project with a missing consent screen now self-heals.
- **Hardened two brittle selectors** on the OAuth-client page (cleanup in the same area):
  - Filter the `arrowDropDownIcon` dropdown arrow to the **visible** element (Google renders several, some hidden).
  - Select the **"Desktop app"** option by visible label instead of the auto-generated Angular id `#_1rif_mat-option-5`, which changes between page versions.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run build` — all clean.
- `npm test` — 547 passed, 2 skipped, 1 todo.
- **Verified end-to-end:** `latchkey auth browser-prepare google-gmail` now completes successfully against a previously-stuck reused project (it self-healed the missing consent screen).

### DIFF
--- a/src/services/google/base.ts
+++ b/src/services/google/base.ts
@@ -345,21 +345,48 @@ interface BrandingResult {
   supportEmail: string;
 }
 
+/**
+ * Configure the OAuth consent screen (branding) for a project.
+ *
+ * Idempotent: returns null when the consent screen is already configured (e.g.
+ * a reused project), so callers can safely invoke it on every run. Google
+ * blocks OAuth-client creation ("you must first configure your consent screen")
+ * until this is done, so it must run before {@link createOAuthClient}.
+ */
 async function configureBranding(
   page: Page,
   projectSlug: string,
   appName: string
-): Promise<BrandingResult> {
+): Promise<BrandingResult | null> {
   await page.goto(`https://console.cloud.google.com/auth/branding?project=${projectSlug}`, {
     timeout: DEFAULT_TIMEOUT_MS,
   });
   const getStartedButton = page.locator('cfc-empty-state-actions .mdc-button__label');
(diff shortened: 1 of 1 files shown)


hynek-urban:
@pseay-imbue Thanks!

PR #92, opened 2026-06-22
Add `latchkey prepare` command and PKCE for Google OAuth

pseay-imbue:
## Summary

Two related changes, both in order to add a more seemless OAuth experience with an official Imbue OAuth Client:

### 1. `latchkey prepare <service> <json>`
A new top-level command that stores a service's credentials directly from a validated JSON payload — the manual equivalent of what `auth browser-prepare` produces, without driving the Cloud Console.

- Services opt in via an optional `prepareFromJson` hook on `Service`, backed by a Zod schema. The base default is "not supported", so every service stays closed until it declares a schema.
- `GoogleService` implements it once: `{ "clientId": "...", "clientSecret": "..." }` → token-less `OAuthCredentials`. Run `auth browser <service>` afterward to complete login (status is `missing` until then, same as `browser-prepare`).
- Strict validation: malformed JSON, missing/unknown fields, or an unsupported service rejects the whole command and stores nothing.
- Forwarded in gateway mode like `browser-prepare`.
- `auth browser-prepare` (Console automation) remains as a fallback.

### 2. PKCE for the Google OAuth login flow
- The authorization request now sends `code_challenge` + `code_challenge_method=S256`, and the token exchange sends the matching `code_verifier`.
- The `client_secret` is still sent (confidential desktop client + PKCE — defense-in-depth). The refresh-token path is unchanged.
- Reuses the existing PKCE helpers in `oauthUtils.ts` (already used by `notion-mcp`); the Google flow simply wasn't calling them.

## Docs
- Updated every Google service's `info` string (surfaced via `services info`) to recommend `prepare` first, with `browser-prepare` as the fallback.
- Updated the README command overview.

## Testing
- `npm run lint`, `npm run typecheck`, `npm run build`, and `npm test` all pass.
- Added coverage for `prepareService` (happy path, overwrite, unknown/unsupported service, malformed JSON, missing/unknown fields, empty fields), the CLI command (success, error exits, gateway forwarding), the gateway endpoint (schema validation + dispatch), and strengthened the PKCE `generateCodeChallenge` test with the RFC 7636 known-answer vector.

## Notes
- The optional-hook design (rather than a required/abstract member) was deliberate: it keeps every existing object-literal `Service` mock compiling and leaves all current services' behavior unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,6 @@ Latchkey is a command-line tool that injects credentials into curl commands.
 	- Open a browser login pop-up window and store the resulting API credentials.
     - This also allows agents to prompt users for credentials.
     - Only some services support this option.
-- `latchkey prepare <service_name> <json>`
-	- Store a service's credentials directly from a validated JSON payload (e.g. an OAuth `{"clientId":"...","clientSecret":"..."}`).
-    - For Google services, `latchkey auth browser-prepare <service_name>` remains the default way to set up an OAuth client; `prepare` is an alternative for supplying an existing client's id/secret directly. Run `latchkey auth browser <service_name>` afterward to complete login.
 
 Latchkey is primarily designed for AI agents. By invoking
 Latchkey, agents can utilize user-provided credentials or prompt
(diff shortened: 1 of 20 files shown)


hynek-urban:
@pseay-imbue Thanks, Preston! I finalized this by doing the following:

- For consistency, I changed `latchkey prepare` to `latchkey auth prepare`.
- Added `auth prepare` support to the `notion-mcp` service.
- Fixed an edge case in Google's `auth browser` implementation that I discovered during manual testing. (This is technically not related to this PR but I just pushed it alongside the other changes.)
- And adjusted the wording of various help texts.

PR #93, opened 2026-06-23
Add Ramp service (browser OAuth PKCE auth + agent-tools API)

pseay-imbue:
Adds a **Ramp** service to latchkey.

- **Browser login (`latchkey auth browser ramp`)**: runs the same OAuth 2.0 authorization-code + PKCE flow as Ramp's official CLI using Ramp's public client, opens Ramp's hosted sign-in/consent screen (which mints an AI agent key), and stores the resulting bearer + refresh token (auto-refreshed). API-client (`set-nocurl`, client_credentials) auth is also supported, incl. `--sandbox`.
- **Agent-tools surface**: a browser/agent-key credential is authorization-level barred from Ramp's standard REST endpoints (`GET /developer/v1/cards` → 403 `DEVELOPER_7077`); it can only use the agent-tools endpoints (`POST /developer/v1/agent-tools/<tool>` with a `rationale` body, even for reads). The service `info` documents this so agents pick the right surface, and `RAMP_OAUTH_SCOPES` is exactly the scope set Ramp's agent-tools OpenAPI declares.
- Bumps the package version to 2.18.0.

Pairs with the detent PR that adds the matching Ramp permission schemas.


src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+ *    only ask for scopes the app already has, and asking for anything else fails
+ *    with `invalid_scope`. So latchkey requests exactly the scopes the user passes
+ *    in -- no more, no less -- which means the minted token can do everything the
+ *    app is allowed to do and nothing it isn't. There is no refresh token in this
+ *    grant; latchkey simply mints a new token with the stored client credentials
+ *    whenever the current one is missing or expired.
+ *
+ * Both credential types reuse existing serializable credential classes
+ * (OAuthCredentials and RampCredentials) so no new credential type has to be
+ * registered. Every API call targets https://api.ramp.com/developer/v1 (or the
```
hynek-urban: This part of the comment doesn't seem to be True :) (There's obviously a new credential class introduced in this PR.)

src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+//
+// latchkey reproduces exactly this flow below, driving the browser step through
+// Playwright (withTempBrowserContext) instead of `open(url)`, and storing the
+// result as OAuthCredentials (which already holds access+refresh tokens and is
+// auto-refreshed by src/apiCredentials/utils.ts). Because the user drives Ramp's
+// own hosted login/consent UI, there are NO scraped selectors in the happy path
+// -- the only Ramp-specific contract is the set of URLs/IDs/scopes constants and
+// the loopback callback shape.
+//
+// !!! MUST BE VALIDATED AGAINST THE LIVE FLOW !!!
```
hynek-urban: If you tested this then I think the section should be removed.

src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+}
+
+export class Ramp extends Service {
+  readonly name = 'ramp';
+  readonly displayName = 'Ramp';
+  // Both the production and sandbox (demo) hosts route here; the stored
+  // credential records which environment it was issued for.
+  readonly baseApiUrls = ['https://api.ramp.com/', 'https://demo-api.ramp.com/'] as const;
+  readonly loginUrl = 'https://app.ramp.com/';
+  readonly info =
```
hynek-urban, line 192: This seems to be overly long. Compare this to the info strings of the established service implementations.

I know that we need to communicate more things here because the setup is more complicated than usual (with the two different surfaces). It would still be good to keep it as concise as possible.

The most important thing that agents need to know after reading this string is where to find the API documentation. Especially the documentation for the agent-tools/ endpoint is quite hidden and not easy to Google to it should be listed here (it's https://api.ramp.com/v1/public/agent-tools/spec/).

src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+  clientSecret: z.string(),
+  scope: z.string(),
+  environment: z.enum(['production', 'sandbox']),
+  accessToken: z.string().optional(),
+  accessTokenExpiresAt: z.string().optional(),
+});
+
+export type RampCredentialsData = z.infer<typeof RampCredentialsSchema>;
+
+export class RampCredentials implements ApiCredentials {
```
hynek-urban: Could we rename this so that it's obvious that it's specific to the "client grant" / nocurl flow?

src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+  readonly objectType = 'ramp' as const;
+  readonly clientId: string;
+  readonly clientSecret: string;
+  readonly scope: string;
```
hynek-urban: Is `scope` ever used for anything? I don't really see it being used anywhere.

src/services/ramp.ts
```diff
@@ -0,0 +1,674 @@
...
+  }
+
+  /**
+   * Browser login: run the ramp-cli OAuth authorization-code + PKCE flow and store
+   * the resulting bearer + refresh token. Always targets production -- the standard
+   * `auth browser` command can't pass an environment, and the sandbox is served by
+   * the existing client_credentials (`set-nocurl ... --sandbox`) pathway instead.
+   */
+  override getSession(appNamePrefix: string): RampOAuthServiceSession {
+    return new RampOAuthServiceSession(this, appNamePrefix, 'production');
```
hynek-urban: The 'production' environment being hardcoded here, I think the following can happen:

- Agent (or user) calls `latchkey auth browser ramp` and logs in to production.
- Agent sees that credentials are valid.
- Agent sends a request to demo-api that fails.

We don't have a good way of supporting "environments" in Latchkey right now. The Ramp service does double-duty but it doesn't really fit the model that well.

I suggest one of the two alternative options instead:

1. Either we could model this as two different "services" inside Latchkey, `ramp` and `ramp-sandbox`. (Having `base.ts` with most of the code and `ramp.ts` and `ramp-sandbox.ts` with the specific overrides - similarly to how google services work.)
2. Or we could drop support for sandbox altogether :) If we need it for testing purposes, we could just leave the replacement constants (for base URL, OAuth client id, etc.) commented out alongside the production ones and, when testing, just uncomment them (and comment-out the production ones).

(I'd personally lean to #2.)

hynek-urban: Another potential reason to drop sandbox support: "Sandbox access is not self-serve. To get a sandbox environment, contact your account manager." (https://support.ramp.com/accessing-the-developer-api/#requesting-sandbox-access)

hynek-urban:
@pseay-imbue Thanks a lot for putting this together! I left you a few comments in the code. Just to double-check: did you manually test this? (At least the agent-tools part.)

src/services/ramp.ts
pseay-imbue: Sg, removed sandbox.

pseay-imbue:
Okay, I've resolved your comments and removed the regular API, so all that's left is the agent tools authentication method, which should now match the accompanying detent PR. I've tested it and it does work :)

hynek-urban:
Awesome, thanks

PR #102, opened 2026-07-18
scripts: add latchkey dev shim that runs the working checkout from source

weishi-imbue:
Ticket: da-0xnk

The `npm link` dev setup fails silently two ways: the global bin serves `dist/`, which goes stale until an explicit rebuild, and it always runs one pinned checkout — even from a worktree or second clone.

`scripts/latchkey`, an opt-in dev shim:

- Resolves the latchkey checkout containing cwd (falling back to the shim's own checkout), so worktrees and second clones just work.
- Runs `src/cli.ts` under node via the checkout's own [tsx](https://tsx.is) (now a devDependency; the docs already prescribe `npx tsx` for the dev scripts) — edits are live, no build step, and the runtime is the same node that users run.
- One code path, read-only: it never executes `dist/`, and never writes to the checkout (`npm install`'s prepare script already generates the gitignored `src/version.ts`). A checkout without tsx or `version.ts` gets a loud error naming the fix, e.g. ``latchkey: no tsx in <root>/node_modules; run `npm install` in <root> first``.
- Install: `npm run install-dev-shim` — symlinks `~/.local/bin/latchkey` and fails with a warning if another install still shadows the shim on PATH.

Published package unchanged (`scripts/` is not in `files`; `bin` still points at `dist/src/cli.js`; tsx is a devDependency only). macOS + Linux; shellcheck-clean.

Plain `node src/cli.ts` is not an option: node's type stripping does not map the codebase's `.js` import specifiers to `.ts` files, which is what tsx adds on top of node.

Tests: `tests/devShim.test.ts` executes the real shim and installer (10 cases covering every branch and error message; no external runtime needed, so all of them run in CI). lint, typecheck, format:check, and the full suite are green.


hynek-urban:
Thanks, @weishi-imbue :) I had issues with Bun in the past and because it's not the primary way of running Latchkey, I wouldn't recommend it as the official way of testing while developing.

I'm not fundamentally opposed to the idea of a "dev shim", though.

But before we do that, is there anything we'd gain by having that over simply ad-hoc calling e.g. `alias latchkey-dev='tsx /path/to/your/latchkey/src/cli.ts'` ? I feel like that fully addresses the mentioned use cases while not requiring any other code or state (maybe just a docs update). Do I miss anything?

PR #103, opened 2026-07-18
auth list: survive corrupt credential store entries

weishi-imbue:
## Problem

One malformed entry in the credential store crashed `latchkey auth list` entirely:

```
ApiCredentialStoreError: Invalid credential data for service databricks: [invalid_union_discriminator]
```

`ApiCredentialStore.getAll()` threw on the first entry that failed schema parsing, so a single stale/corrupt entry hid every other credential — exactly when the listing is needed to find the bad entry — and the error named no remedy. `services list --viable` shares `getAll()` and crashed the same way.

## Fix

The parse-tolerance lives at the store layer, since both readers of `getAll()` are read-only diagnostics and schema/version skew belongs at the deserialization layer:

- **`getAll()`** now parses each entry independently and returns `{ credentials, brokenEntries }`. Broken entries carry the claimed `objectType` (the key diagnostic under version skew) and a compact per-field parse error. Nothing is swallowed — every failure is surfaced to the caller.
- **`authList`** emits a broken entry as a visible row: `credentialStatus: "corrupt"` (new `ApiCredentialStatus` member), the claimed type, and an `error` naming the remedy. Rows are built through a `Map` so a store key like `__proto__` cannot vanish via the object prototype setter.
- **`services list --viable`** treats a corrupt credential as non-viable (it cannot be injected into a request) instead of crashing; `auth list` is the diagnostic that flags it.
- **`auth clear <service>`** now deletes by store key before consulting the registry. A corrupt entry commonly belongs to a service the running version does not know, and the registry-first check rejected exactly those names with "Unknown service" — making the advertised remedy impossible to run. Unknown names with nothing stored still fail as before.
- **`get()`** (single-service reads) still throws loudly, but its error now names the same remedy.

Note: this changes the return shape of the public `getAll()`; the new `BrokenCredentialEntry` / `CredentialStoreListing` types are exported from the package index. Both in-repo callers are updated; external callers get a compile error rather than a behavior change.

`services info` on a corrupt entry still crashes via the `get()` path (pre-existing); left out to keep this PR single-purpose.

## Tests

Real malformed fixtures (no mocked parse errors) at all four layers: store unit tests, `authList`/`servicesList` operation tests, CLI command test, and gateway endpoint test. Nine of the new tests fail on the code before this change. One fixture orders the corrupt entry before a valid one so a stop-at-first-corrupt regression cannot pass. A CLI test seeds a corrupt entry for an unregistered service and asserts the advertised `auth clear` remedy actually removes it.

Verified end-to-end with the built CLI: `auth list` over a store containing a corrupt entry exits 0, lists the valid credentials, flags the corrupt one with the remedy; running the remedy removes only that entry.

### DIFF
--- a/src/apiCredentials/base.ts
+++ b/src/apiCredentials/base.ts
@@ -9,6 +9,8 @@ export enum ApiCredentialStatus {
   Valid = 'valid',
   Invalid = 'invalid',
   Unknown = 'unknown',
+  /** The stored credential data does not match any known credential schema. */
+  Corrupt = 'corrupt',
 }
 
 /**
--- a/src/apiCredentials/store.ts
+++ b/src/apiCredentials/store.ts
@@ -21,7 +21,7 @@ export class ApiCredentialStoreError extends Error {
 type StoreData = Record<string, unknown>;
 
 /** A store entry whose data does not match any known credential schema. */
-export interface BrokenCredentialEntry {
+export interface CorruptCredentialEntry {
   /** The entry's claimed objectType, or null when it is missing or not a string. */
   readonly objectType: string | null;
   readonly error: string;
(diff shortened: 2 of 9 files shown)


hynek-urban:
@weishi-imbue Thanks for the PR :) How did the corrupt credentials end up in the store?

(I'm currently changing the structure of the credentials store in order to support multiple credentials - I may replicate the logic from this PR in the new code but I'd like to understand how it happened in the first place.)

PR #111, opened 2026-07-29
Support generic browser login flows for registered services

qi-imbue:
## What

Registered (custom) services can now have browser automation attached to them. `latchkey services register` gains two options that pick one of Latchkey's generic login flows and configure it:

```bash
latchkey services register my-intranet \
  --base-api-url="https://intranet.example.com/api/" \
  --login-url="https://intranet.example.com/login" \
  --login-flow=cookie-capture \
  --login-flow-params='{"cookieKeys": ["sessionid", "csrftoken"]}'

latchkey auth browser my-intranet                    # opens the login page, waits for the cookies
latchkey curl https://intranet.example.com/api/me    # sends Cookie: sessionid=…; csrftoken=…
```

One flow ships, `cookie-capture`: it opens the login URL and watches the `Set-Cookie` headers of the responses that arrive while the user signs in. Its parameters are `cookieKeys` (required, at least one) and `cookieUrl` (optional, defaults to the login URL). `latchkey services register --help` documents both, and the flow list there is generated from the registry, so a flow added later documents itself.

Behavioral details worth knowing:

- **All captured cookies go into one header** — `Cookie: sessionid=<value>; csrftoken=<value>`. That is the only portable form: RFC 6265 has the user agent send a single `Cookie` header, and servers that see several header lines commonly read just the first (checked against curl — two `-H "Cookie: …"` arguments put two lines on the wire).
- **A name matching more than once keeps every match**, since the same name can be set both host-only and domain-wide, and a browser sends both.
- **Header handling follows RFC 6265** for `Domain`, `Path`, `Max-Age` and `Expires`: a cookie counts only if a browser would send it to the configured URL, a cleared cookie is dropped again, and a re-issued cookie replaces the one in the same scope.
- **Known limitation:** because capture is response-driven, a cookie set purely from page script (`document.cookie`) is never seen, and neither is one that an already-signed-in browser session does not re-issue. Both leave `auth browser` waiting until the user closes the window. This is verified, not theoretical — see Testing. Reading Playwright's own cookie jar (`context.cookies(url)`) instead would close the gap, at the cost of that API's quirks; `cookieUtils` records the trade-off.

`--login-flow` cannot be combined with `--service-family`, since a family already brings its own login flow and credential shape.

## Why

Registered services previously supported only `latchkey auth set`. Many internal and self-hosted services authenticate API calls with a plain session cookie, leaving users to copy it out of their browser by hand.

The flags are deliberately about *flows* rather than cookies: naming the automation and passing its parameters as JSON means a second flow — or a pluggable one — can be added without touching the CLI surface.

## How

Four new modules, and one change to shared machinery:

- `src/services/core/loginFlow.ts` — what a flow *is*. A **flow class** is a kind of automation, carrying `flowName`, `summary`, `details` and a Zod schema; a **flow** is one configuration of that kind, which describes itself and creates sessions; a **session** is a single run. Two types express this, not one, because the metadata is listed before any flow is configured — so it must be static, and TypeScript has no `abstract static`. Every flow's constructor takes `unknown` and reads its parameters through `parseLoginFlowParams`, which infers their type from the flow's own schema: a flow whose stored parameters disagree with the schema it declared does not compile.
- `src/services/core/loginFlowRegistry.ts` — which flows *exist*. Because constructors are uniform, the registry is a plain array of the flow classes and `resolveLoginFlow` is a lookup and a `new` — no generic bridge, no cast. Each flow states its own obligation with `satisfies LoginFlowClass`, so a missing static errors on the flow rather than here.
- `src/services/core/cookieCapture.ts` — the `cookie-capture` flow and its session. The session is not exported; the only exported class is the flow.
- `src/cookieUtils.ts` — `CookieJar`: `Set-Cookie` headers go in, and it answers which cookies would be sent to a given URL. Parsing, scope matching and cookie identity are internals of the jar.
- `src/services/core/base.ts` — `SimpleServiceSession` gains a `capturedCredentials` getter, and `onResponse` returns its promise (the abstract signature widens to `void | Promise<void>`; the login loop discards it explicitly and still polls). **This affects every service**, and pays off immediately: `tests/servicesAgainstRecordings.test.ts` no longer casts to read a protected field, and no longer sleeps 10 ms per replayed response.
- `src/services/core/registered.ts` — `RegisteredService` takes an options object rather than positional parameters, typed as a **union of the two mutually exclusive login sources** (a family service, or a flow plus the URL it starts from). The union needs `never` markers to hold: a plain `A | B` still accepts an object carrying both, since excess property checking against a union permits any property known to some member. A side effect is that the options type no longer overlaps `Service` structurally, so passing a family service positionally is now an error rather than being read as "options with a loginUrl".
- `src/configDataStore.ts`, `src/serviceRegistry.ts`, `src/cliCommands.ts` — persistence of the `loginFlow: { name, params }` entry, resolution on load, CLI flags and validation. A stored flow that no longer resolves (config from a newer Latchkey, or hand-edited) costs the service its browser login but leaves the service usable with `auth set`.

The two login-flow modules are split rather than one, and the types live there rather than in `base.ts`, because a flow importing the module that lists the flows is a cycle whose evaluation order decides whether it throws. Flows import `loginFlow`; only the registry imports the flows.

## Testing

- `tests/cookieCapture.test.ts` drives the public path a real login takes — a flow, the service it is registered on, the session that service hands out, and the credentials that session reports — with responses fed in by hand instead of by a browser. No class or helper is exported for the tests' benefit; the one test-facing addition is the `capturedCredentials` getter, noted below.
- `tests/cookieUtils.test.ts` covers the jar: path and domain defaults, explicit `Domain`/`Path`, subdomains, the three ways a cookie is cleared, re-issue in the same scope versus a second scope, and malformed headers.
- CLI tests cover registering with a flow, config round-tripping, unknown flow, malformed JSON, schema-invalid parameters, every rejected flag combination, and a service whose stored flow no longer resolves. Invalid `RegisteredService` option combinations are asserted at compile time with `@ts-expect-error`, which the suite enforces by running `tsc --noEmit`.
- Verified end-to-end against real Chrome and a local fake service whose sign-in sets `sessionid` and `csrftoken` on staggered XHRs: the login waited for both, and the API received exactly one `Cookie` line — `{"cookieHeaderLines":1,"receivedCookie":"sessionid=s3cr3t-value; csrftoken=csrf-value"}`.
- Verified the known limitation the same way: pointed at a page that sets its cookie with `document.cookie`, `auth browser` was still waiting after 14 seconds.
- `npm run lint`, `npm run typecheck`, `npm test` (752 tests) all pass.

## Manual testing

- Register a real cookie-authenticated service, run `latchkey auth browser <name>`, sign in, and confirm the window closes on its own and `latchkey curl` reaches an authenticated endpoint.
- A service needing several cookies: confirm the login does **not** finish until the last one is set.
- A login that redirects through an SSO domain, with `cookieUrl` naming the API host.
- A service whose session cookie is set by page script rather than by a response — expected to hang; worth confirming the failure mode is tolerable before this ships.
- Close the browser window mid-login: should report login cancelled, not hang or store partial credentials.
- Sign-in flows that set a cookie, clear it, and set it again — the final value should be the stored one.
- Error paths: `--login-flow=typo` lists the available flows; `--login-flow-params='{bad json}'` and `'{"cookieKeys": []}'` are rejected at registration with a readable message.
- `latchkey services register --help` reads correctly at 80 columns.
- **Existing browser logins still work** — the `onResponse` change touches all of them. Worth exercising one built-in login (Slack or Google) end to end, and replaying any local recordings through `tests/servicesAgainstRecordings.test.ts`, which has no recordings in this repo and therefore skips in CI.

## Notes for review

- `--login-url` is required with `--login-flow`, and also with a `--service-family` whose family has a browser login; it is rejected otherwise. That is why it stays a top-level flag rather than becoming a flow parameter.
- `cookieUrl` is validated as a URL by the flow's schema, at registration time. `--login-url` still is not, though a bad value fails visibly when the browser tries to open it.
- The `RegisteredService` constructor repeats the exclusion checks its options type already enforces, because latchkey is consumed as a library from plain JavaScript too. Tests cover both layers.
- `SimpleServiceSession.capturedCredentials` is public for callers that drive a session from responses they obtained themselves; today the only such callers are the two test files, and its docstring says so. Removing it means either casting past `protected` again, or driving `login()` with a real browser — which CI has no way to run, so that test would skip rather than cover anything.

## Open design questions (deliberately left as-is)

- No in-browser guidance: the login page just opens, and the window closes itself once the cookies are set.
- No timeout: waits indefinitely, like the existing built-in logins. Closing the window surfaces as "login cancelled".


src/cliCommands.ts
```diff
@@ -445,10 +445,25 @@ export function registerCommands(program: Command, deps: CliDependencies): void
       'Name of the built-in service to use as a template, if any (e.g. gitlab)'
     )
     .option('--login-url <url>', 'Login URL for browser-based authentication, if applicable')
+    .option(
```
qi-imbue, line 527: I don't like the flag design here - the flags are not really orthogonal to each other - but I'm not sure what's a better way. Thoughts?

Claude suggested that if in future we want to support multiple "flavours" of generic login flows, we can introduce a `--login-flow` flag, so for a cookie-capturing custom service you'll need `--login-flow=cookie-capture --cookie-key=...`.

src/services/core/base.ts
```diff
@@ -314,11 +314,20 @@ export abstract class ServiceSession {
...
+   * implement this to inspect the live page.
+   */
+  protected checkLoginProgress?(page: Page): Promise<void>;
+
   /**
    * Wait until the browser login phase is complete.
    */
   private async waitForLoginComplete(page: Page): Promise<void> {
     while (!this.isLoginComplete()) {
+      await this.checkLoginProgress?.(page);
```
hynek-urban: Do we actually need this? Presumably, a cookie needs to be set first which would typically happen through a response. In fact, listening for the Set-Cookie header would be more robust in the sense that we wouldn't accidentally register an obsolete cookie.

(The main downside being if the webpage actually sets the cookie via javascript. Is that the case for the service you're currently interested in? If so then that would probably be a reason to keep it like this.)

src/cliCommands.ts
hynek-urban: I agree with this. I would maybe go even further and declare `--login-flow=cookie-capture` alongside `--login-flow-params='{"cookieKey": "abcdef"}'`. That way we'll be prepared for a future mechanism where custom login flows could be supplied by plugins / extensions.

This would imply having a concept of a `LoginFlow` and `LoginFlowParams` (or something similar) and have it be present in the interfaces and in the registry / serialized config (as opposed to the "naked" cookie-related params).

What do you think?

README.md
```diff
@@ -221,7 +221,63 @@ latchkey auth set mastodon -H "Authorization: Bearer <your_access_token>"
 latchkey curl https://mastodon.social/api/v1/timelines/public?limit=2
 ```
 
-User-registered services only support authentication via static curl arguments provided through `latchkey auth set`.
```
hynek-urban, line 224: I'm trying to keep the README human-written :)

It's also too long already so for each new feature, I've been trying to be very brief. (Eventually, I'd like Latchkey to have better docs and not just squeeze everything into the README, but I didn't get to it yet.)

Would it be possible to keep it short here - just a single sentence or two, pointing at `latchkey services register --help` in the CLI and making sure the `--help` invocation is actually helpful (while still not as verbose as this version :) )

src/cliCommands.ts
hynek-urban: (Or even just `--login-flow='{"type": "cookieCapture", "cookieKey": "abcdef"}'`, not sure.)

src/services/core/base.ts
qi-imbue: Hmm good point. I've seen a couple of articles as well as Claude saying that auth cookies should always be set by the server as an HttpOnly cookie - the reason is to mitigate XSS attacks - and I find that quite believable. So for auth cookies it's probably pretty rare to have them set in JavaScript.

I also manually tested claude.ai's and the auth cookie indeed set by the server as an HttpOnly cookie.

README.md
qi-imbue: Ack, will do

src/cliCommands.ts
qi-imbue: I slightly prefer two flags, going with that

src/cookieUtils.ts
qi-imbue, line 1: Switching to capturing Set-Cookie headers meant that we end up reimplementing a cookie jar. If we don't like that, we can

- Introduce a dependency that already implements these
- Adopt a hybrid approach where we trigger an update to the cookie jar on HTTP response headers, but still use Playwright to actually read the cookie jar (this can be slightly tricky if there's some detail between the HTTP response and Chromium's cookie jar update though)

src/services/core/base.ts
```diff
@@ -293,8 +300,13 @@ export abstract class ServiceSession {
...
   /**
    * Handle a response during the headful login phase.
+   *
+   * Sessions that do asynchronous work here return a promise, so that a caller
+   * driving responses in on its own — replaying a recording, say — can wait for
+   * one to be processed before sending the next. The login itself does not
+   * wait: it polls {@link isLoginComplete} instead.
    */
-  abstract onResponse(response: Response): void;
+  abstract onResponse(response: Response): void | Promise<void>;
```
qi-imbue, line 309: The Promise<void> return was added to make tests work

src/services/core/base.ts
```diff
@@ -414,19 +426,36 @@ export abstract class ServiceSession {
...
+   * still going.
+   *
+   * Exists for the tests, and they are its only callers: a real login goes
+   * through {@link login}, which returns the credentials once the phase
+   * completes. It is public rather than protected because the alternative is
+   * what the tests did before — casting past `protected` to read the field —
+   * which typechecks forever and quietly stops meaning anything the moment the
+   * internals change.
+   */
+  get capturedCredentials(): ApiCredentials | null {
```
qi-imbue, line 440: Same, to make unit tests work.

src/services/core/base.ts
hynek-urban: Right. As far as I'm concerned, we could just always return `Promise<void>`.

src/cookieUtils.ts
hynek-urban: I think it's fine to keep this implementation :) I'd just love to slightly adjust the file organization:

- Could we move the contents of `cookieUtils.ts` into `cookieCapture.ts` and make it private? That way we keep a "lid" on the complexity - it's all hidden behind the LoginFlow interface and unless you're specifically interested in that one particular login flow, you never need to care about the cookie implementation details in any way. (If we ever need cookie handling more generally, we can factor it out back to a cookieUtils module again.)
- Could we group the login flow related modules more closely together? For example:
  - `services/core/loginFlows/base.ts`
  - `services/core/loginFlows/registry.ts`
  - `services/core/loginFlows/cookieCapture.ts`
  - (The goal is similar to the previous point - making it obvious that you don't need to care about `cookieCapture.ts` unless you're interested in login flows.)

hynek-urban (commented):
Thanks, @qi-imbue , this is great :) Eventually, I think even some built-in services will be able to use the login flow structure.

Anyway, I left you a few comments. Once you're happy with everything, let me know and I'll merge and publish this so that it can be used downstream!

src/services/core/base.ts
qi-imbue: Yeah, that would require changing all the implementations though, so I opted for the slightly ugly union return type.

PR #113, opened 2026-07-31
Add ngrok as a built-in service

boweiliu:
## Summary

Adds **ngrok** as a built-in latchkey service.

ngrok issues personal API keys from its dashboard (shown once), mirroring the existing **Linear** connector: sign the user in via a headful browser, create a fresh API key, and read its value from the one-time reveal dialog.

The credential is stored as raw curl arguments that inject **both** the bearer API key and the mandatory `ngrok-version: 2` header (per the [ngrok API docs](https://ngrok.com/docs/api/#authentication)), so a plain `latchkey curl https://api.ngrok.com/...` works without the caller remembering the version header.

## Verification (done against real ngrok)

- Selectors/flow confirmed by driving a live ngrok dashboard login: `/api-keys/new` opens the create dialog directly; description field prefilled (cleared before typing); key revealed once in the success dialog's `<pre>`.
- Login-completion detected from the post-auth dashboard document navigation (works for password and SSO).
- A minted key stored via `latchkey auth set` → `latchkey curl https://api.ngrok.com/api_keys` returns **200**.
- End-to-end tunnel: the key minted an agent authtoken via `POST /credentials` (201), which the `ngrok` CLI used to open a working tunnel (public URL forwarded to a local server, 200 through the tunnel).

## Note: API key vs authtoken

The stored credential is a REST **API key**. The `ngrok` agent (CLI/SDK/K8s) uses an **authtoken** — a different credential. `docs/ngrok.md` documents deriving an authtoken on demand via `POST /credentials`.

## Files
- `src/services/ngrok.ts` — the connector
- `src/services/index.ts`, `src/serviceRegistry.ts` — registration
- `docs/ngrok.md` — usage + tunnel authtoken flow

## Test coverage note (recordings-replay)

ngrok is added to the `BLACKLIST` in `tests/servicesAgainstRecordings.test.ts`. That test replays a recorded login session and extracts the API credential from the recorded network traffic. ngrok is a `BrowserFollowupServiceSession`: its API key is minted by an interactive browser action and read from a one-time reveal dialog, so it is **never present in the recorded login traffic** — replay cannot reconstruct it. This is the same reason `linear` / `dropbox` / `github` are excluded. (Recordings themselves are gitignored, so the entry is also a forward-looking guard for anyone who records an ngrok session locally.) See the PR comment below for the full reasoning, including the one pre-existing gap (`todoist`).


boweiliu:
https://github.com/imbue-ai/mngr-internal/pull/235

boweiliu:
Published the companion **dev workbench** as a public inspiration: https://github.com/boweiliu/latchkey-dev-workbench

It's the toolkit + playbook this connector was built with -- the bridge shell channel, `deskrun`, the self-restart loop, the patch tools, the ngrok connector, and the full build journal (including the schema-before-grant gotcha and the uv-cache hot-mod trick). ngrok is the worked example, taken end-to-end and restart-durable on a real Minds gateway.

hynek-urban:
Thanks, @boweiliu , this is useful!

Could you please add support for `ngrok` also to Detent? https://github.com/imbue-ai/detent

If you create a PR over there, I will then publish a new version of Detent, then I'll merge this PR, bump the Detent dependency version and will publish a new version of Latchkey with ngrok support, yay :)

boweiliu:
Detent side (the ngrok-api scope schema) is up: https://github.com/imbue-ai/detent/pull/22 — the piece needed before this can merge.

boweiliu:
### Recordings-blacklist decision (resolved): ngrok belongs in the blacklist

**Decision:** ngrok is added to `BLACKLIST` in `tests/servicesAgainstRecordings.test.ts`.

**Mechanism.** `discoverRecordings()` iterates `scripts/recordings/<svc>/`, and tests any service that has a committed `login_session.json` and is not blacklisted. The test extracts the API credential from the **recorded network traffic** of a login and checks the connector against it.

**Why ngrok can't be recording-tested.** ngrok is a `BrowserFollowupServiceSession`: after sign-in, its API key is created and read from the dashboard's one-time "New API Key" reveal dialog (a DOM action), so the key is **not in the recorded login network traffic**. Replaying a recording therefore cannot reconstruct a working credential — the same failure mode as the other browser-followup services already excluded.

**Consistency check (evidence).** Session type per connector:

| connector | session | blacklisted? |
|---|---|---|
| dropbox | `BrowserFollowupServiceSession` | yes |
| github | `BrowserFollowupServiceSession` | yes |
| linear | `BrowserFollowupServiceSession` | yes |
| **ngrok** | `BrowserFollowupServiceSession` | **now yes** |
| todoist | `BrowserFollowupServiceSession` | no (pre-existing gap) |
| slack, discord | `SimpleServiceSession` (creds in traffic) | no (correct) |

ngrok is the exact analog of `linear`, so excluding it is the consistent call.

**Course-correction, in the open.** An earlier commit on this branch added ngrok to the blacklist, then a follow-up reverted it (reasoning: recordings are gitignored, and `todoist` is also a followup service yet unlisted). On closer analysis that revert was wrong: the blacklist's stated criterion is "services that require special followup steps," which ngrok plainly does, and `linear` — its exact pattern — is listed. `todoist` is a pre-existing gap, not a model to copy; it arguably should be blacklisted too, but that's left out of scope here to keep this change focused. Net: ngrok is (re-)added, with the criterion documented inline.

PR #117, opened 2026-08-07
Add Hugging Face connector

boweiliu:
Adds a Hugging Face service connector: sign-in mints a read-only User Access Token (revealed once, scraped from the reveal dialog, like Linear); stored as a bearer credential; matches the Hub (`huggingface.co`) and the Inference router (`router.huggingface.co`). `getAccount` reports the username via `/api/whoami-v2`. Listed in SKILL.md; blacklisted from `servicesAgainstRecordings` (browser-followup service).

## Usage (scopes, defined in imbue-ai/detent#23)
- **Download models/datasets:** `huggingface-api` + `huggingface-read` (safe methods incl. the HEAD before a download).
- **Run inference:** `huggingface-api` + `huggingface-inference` (POST to the router).
- **Create repos / upload:** `huggingface-api` + `huggingface-write` (Hub writes only).

## Verified live
Browser login minted a sealed read-only token; downloads work (`GET`/`HEAD resolve` -> 200), writes are denied under a read grant (403), inference reaches the router. Token never leaves the gateway. lint/typecheck/format/`npm test` (758) green.

Needs a Detent release + dependency bump to ship (as ngrok did in #113).

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -77,16 +77,13 @@ Every service must include an `info` property that provides developer notes. Thi
 - Special requirements (e.g., need to run `latchkey auth browser-prepare` first)
 - Any caveats or limitations
 
-Keep the `info` text about the *service* itself, not the Detent scopes or
-permission names an agent might request. Latchkey is intentionally fairly
-standalone: it matches URLs and injects credentials, and it does not know which
-scopes a service exposes. Scopes and permissions are defined in
-[Detent](https://github.com/imbue-ai/detent) (the enforcement layer), which is
-aware of Latchkey and is where those names belong. Pointing an agent at the
-service's own documentation (e.g. an `llms.txt` index) lets it discover the right
-endpoint and capability for the task instead of baking Detent's vocabulary into
-the connector. The same applies to comments on the `Service` class: describe what
-the token can do, not which Detent permissions it backs.
+Keep `info` about the service itself -- links to its own docs, caveats, special
+requirements -- not the Detent scopes or permission names an agent might request.
+Latchkey only matches URLs and injects credentials; scopes live in
+[Detent](https://github.com/imbue-ai/detent), the enforcement layer, which is the
+side that knows about Latchkey. Point agents at the service's own documentation
+(e.g. an `llms.txt` index) rather than baking Detent's vocabulary into the
+connector. The same applies to comments on the `Service` class.
 
 ### Potentially useful helpers
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -77,6 +77,17 @@ Every service must include an `info` property that provides developer notes. Thi
 - Special requirements (e.g., need to run `latchkey auth browser-prepare` first)
 - Any caveats or limitations
 
+Keep the `info` text about the *service* itself, not the Detent scopes or
+permission names an agent might request. Latchkey is intentionally fairly
+standalone: it matches URLs and injects credentials, and it does not know which
+scopes a service exposes. Scopes and permissions are defined in
+[Detent](https://github.com/imbue-ai/detent) (the enforcement layer), which is
+aware of Latchkey and is where those names belong. Pointing an agent at the
+service's own documentation (e.g. an `llms.txt` index) lets it discover the right
+endpoint and capability for the task instead of baking Detent's vocabulary into
+the connector. The same applies to comments on the `Service` class: describe what
+the token can do, not which Detent permissions it backs.
+
 ### Potentially useful helpers
 
 #### Codegen
--- a/docs/development.md
+++ b/docs/development.md
@@ -78,9 +78,7 @@ Every service must include an `info` property that provides developer notes. Thi
 - Any caveats or limitations
 
 Keep `info` (and comments on the `Service` class) about the service itself, not
-the Detent scopes or permission names an agent might request. Scopes live in
-[Detent](https://github.com/imbue-ai/detent), which is the side that knows about
-Latchkey -- not the other way around.
+the Detent scopes or permission names (not everyone uses them).
 
 ### Potentially useful helpers
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -77,11 +77,10 @@ Every service must include an `info` property that provides developer notes. Thi
 - Special requirements (e.g., need to run `latchkey auth browser-prepare` first)
 - Any caveats or limitations
 
-Keep `info` about the service itself -- links to its own docs, caveats, special
-requirements -- not the Detent scopes or permission names an agent might
-request. Latchkey only matches URLs and injects credentials; scopes live in
+Keep `info` (and comments on the `Service` class) about the service itself, not
+the Detent scopes or permission names an agent might request. Scopes live in
 [Detent](https://github.com/imbue-ai/detent), which is the side that knows about
-Latchkey. The same applies to comments on the `Service` class.
+Latchkey -- not the other way around.
 
 ### Potentially useful helpers
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -78,12 +78,10 @@ Every service must include an `info` property that provides developer notes. Thi
 - Any caveats or limitations
 
 Keep `info` about the service itself -- links to its own docs, caveats, special
-requirements -- not the Detent scopes or permission names an agent might request.
-Latchkey only matches URLs and injects credentials; scopes live in
-[Detent](https://github.com/imbue-ai/detent), the enforcement layer, which is the
-side that knows about Latchkey. Point agents at the service's own documentation
-(e.g. an `llms.txt` index) rather than baking Detent's vocabulary into the
-connector. The same applies to comments on the `Service` class.
+requirements -- not the Detent scopes or permission names an agent might
+request. Latchkey only matches URLs and injects credentials; scopes live in
+[Detent](https://github.com/imbue-ai/detent), which is the side that knows about
+Latchkey. The same applies to comments on the `Service` class.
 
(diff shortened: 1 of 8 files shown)


src/services/huggingface.ts
```diff
@@ -0,0 +1,155 @@
...
+// Hub API + Inference router. Hub model/dataset downloads and the whoami check
+// live under huggingface.co; hosted inference goes through the router.
+const HF_HUB_BASE_URL = 'https://huggingface.co/';
+const HF_ROUTER_BASE_URL = 'https://router.huggingface.co/';
+
+const HF_LOGIN_URL = 'https://huggingface.co/login';
+
+// Navigating here opens the "Create new Access Token" form preset to a
+// read-only token; the value is shown once in a dialog after creation.
+const HF_NEW_READ_TOKEN_URL = 'https://huggingface.co/settings/tokens/new?tokenType=read';
```
hynek-urban: Why is it a read-only token? That probably wouldn't play well with the "-write-all" Detent permission, among others.

src/services/huggingface.ts
```diff
@@ -0,0 +1,155 @@
...
+  readonly name = 'huggingface';
+  readonly displayName = 'Hugging Face';
+  readonly baseApiUrls = [HF_HUB_BASE_URL, HF_ROUTER_BASE_URL] as const;
+  readonly loginUrl = HF_LOGIN_URL;
+  // Signing in mints a read-only User Access Token. To mint and run hosted
+  // inference, request the `huggingface-api` scope with `huggingface-inference`
+  // (add `huggingface-read` to list/download models & datasets; `huggingface-write`
+  // covers creating repos and uploading).
+  readonly info =
+    'Hugging Face Hub + Inference. Signing in mints a read-only token. Request the ' +
```
hynek-urban: 1. Unless it's justified, let's drop the "read-only token" sentence.
2. Can we point agents to the documentation here? https://huggingface.co/docs/hub/llms.txt
3. Let's drop the mentions of the scopes and permissions (they're only loosely connected to Latchkey).

hynek-urban (commented):
@boweiliu Thanks a lot! My main question is: is it intentional that the token is read-only?

src/services/huggingface.ts
boweiliu: ah yeah good point. i was focused here on the ability to download models and run inference (both of which are under the "read" generic token) but i forgot some folks use hugging face pretty much like github, to upload models too. will take a look here.

src/services/huggingface.ts
boweiliu: 1. yup will follow up from the other comment
2. will do
3. good point, i'll also add some DEVELOPING.md notes here to make sure authors and agents are guided through this in future (basically: detent should be aware of latchkey but latchkey should be fairly standalone).

src/services/huggingface.ts
boweiliu: huggingface tokens actually have quite a lot of fine grained ness. if we want inference only that is a setting. if we want model weights cloning only that's also a thing. will look more into docs

boweiliu: Done in 1a99583. Switched the browser login to `tokenType=write` — a classic HF write token grants read and write access to the user's repositories and also serves as a bearer token for the Inference Providers router, so one minted token backs every HF permission an agent might be granted (read, write, inference). Detent still enforces the per-grant scope; the token is only the capability ceiling.

src/services/huggingface.ts
boweiliu: All three done in 1a99583:

1. Dropped the "read-only token" sentence (no longer applicable — we now mint a write token).
2. Pointed agents at the Hub docs index `https://huggingface.co/docs/hub/llms.txt` in the `info` text.
3. Dropped the Detent scope/permission names from the connector's `info` (they're Detent's concern, only loosely connected to Latchkey). Added a short note to `docs/development.md` so future connectors keep Latchkey standalone — Detent knows about Latchkey, not the reverse. The matching Detent-side note is imbue-ai/detent#26.

tests/servicesAgainstRecordings.test.ts
```diff
@@ -43,7 +43,7 @@ const DEFAULT_RECORDING_NAME = 'login_session.json';
 // from the dashboard's "New API Key" dialog, exactly like linear).
 // NOTE: todoist is also a BrowserFollowupServiceSession and arguably belongs here
 // too; it predates this list and is left as-is to keep this change scoped.
-const BLACKLIST = new Set(['dropbox', 'github', 'linear', 'ngrok']);
+const BLACKLIST = new Set(['dropbox', 'github', 'linear', 'ngrok', 'huggingface']);
```
boweiliu, line 46: remind me of the justification for this?

src/services/huggingface.ts
```diff
@@ -0,0 +1,162 @@
...
+export class Huggingface extends Service {
+  readonly name = 'huggingface';
+  readonly displayName = 'Hugging Face';
+  readonly baseApiUrls = [HF_HUB_BASE_URL, HF_ROUTER_BASE_URL] as const;
+  readonly loginUrl = HF_LOGIN_URL;
+  // The `info` text is what an agent reads to decide whether this service is
+  // useful. Keep it about the *service*, not the Detent scopes/permissions --
+  // those are Detent's concern and only loosely connected to Latchkey (see
+  // DEVELOPING.md). Point agents at the docs index so they can find the right
+  // page for the task they have in mind.
```
boweiliu: is this comment really necessary? it's what i told you, not what code readers need to know

src/services/huggingface.ts
```diff
@@ -0,0 +1,162 @@
...
+// live under huggingface.co; hosted inference goes through the router.
+const HF_HUB_BASE_URL = 'https://huggingface.co/';
+const HF_ROUTER_BASE_URL = 'https://router.huggingface.co/';
+
+const HF_LOGIN_URL = 'https://huggingface.co/login';
+
+// Navigating here opens the "Create new Access Token" form preset to a classic
+// write token (read + write, and a valid bearer for inference); the value is
+// shown once in a dialog after creation.
+const HF_NEW_TOKEN_URL = 'https://huggingface.co/settings/tokens/new?tokenType=write';
```
boweiliu, line 59: theres a bit of rough ux here cuz navigating to this page sometimes triggers the github style "sudo" protection and rerequests pw, but it's in the wrong latchkey tab so users have to manually swap voer

docs/development.md
```diff
@@ -77,6 +77,17 @@ Every service must include an `info` property that provides developer notes. Thi
...
+Keep the `info` text about the *service* itself, not the Detent scopes or
+permission names an agent might request. Latchkey is intentionally fairly
+standalone: it matches URLs and injects credentials, and it does not know which
+scopes a service exposes. Scopes and permissions are defined in
+[Detent](https://github.com/imbue-ai/detent) (the enforcement layer), which is
+aware of Latchkey and is where those names belong. Pointing an agent at the
+service's own documentation (e.g. an `llms.txt` index) lets it discover the right
+endpoint and capability for the task instead of baking Detent's vocabulary into
+the connector. The same applies to comments on the `Service` class: describe what
+the token can do, not which Detent permissions it backs.
```
boweiliu: could this be more succinct and more generally helpful? sounds too mucch like we're responding directly to the PR comments made here. read teh rest of the doc for tone and succinctness. style matters.

tests/servicesAgainstRecordings.test.ts
boweiliu: Same as ngrok/linear: huggingface is a `BrowserFollowupServiceSession` whose token is minted by the one-time reveal dialog, so it is not present in the recorded login network traffic and a recording replay cannot reconstruct it. Spelled that out in the comment in 9cc5b27.

src/services/huggingface.ts
boweiliu: Agreed, removed in 9cc5b27. The guidance lives in `docs/development.md` now; the connector does not need to restate it.

src/services/huggingface.ts
boweiliu: Yeah, the `/settings/tokens` page is behind HF's sudo re-auth. We can't stop HF asking for the password, but the wrong-tab part is fixable: I added `page.bringToFront()` before the `goto` in 9cc5b27 so the prompt surfaces on the tab the user is already on. The sudo prompt itself is HF's behavior and out of our hands.

docs/development.md
boweiliu: Rewrote it in 9cc5b27 -- shorter, in the doc's own voice, and stated as general guidance rather than a reply to the review.

docs/development.md
```diff
@@ -77,6 +77,14 @@ Every service must include an `info` property that provides developer notes. Thi
 - Special requirements (e.g., need to run `latchkey auth browser-prepare` first)
 - Any caveats or limitations
 
+Keep `info` about the service itself -- links to its own docs, caveats, special
+requirements -- not the Detent scopes or permission names an agent might request.
+Latchkey only matches URLs and injects credentials; scopes live in
+[Detent](https://github.com/imbue-ai/detent), the enforcement layer, which is the
+side that knows about Latchkey. Point agents at the service's own documentation
+(e.g. an `llms.txt` index) rather than baking Detent's vocabulary into the
+connector. The same applies to comments on the `Service` class.
```
boweiliu: still verbose, reads like the same idea 3 times. improve

boweiliu: Trimmed again in e5ae103 -- one statement of each idea.

PR #119, opened 2026-08-10
Add an OpenHost self-hosted service connector

boweiliu:
## What

Adds an **OpenHost** ([imbue-openhost/openhost](https://github.com/imbue-openhost/openhost)) service connector for self-hosted instances.

OpenHost is always self-hosted, so there is no canonical API host to pin. The connector is a **family** with no URL of its own; a user registers their own instance against it:

```
latchkey services register my-openhost \
  --service-family openhost \
  --base-api-url "https://oh.example.com/" \
  --login-url "https://oh.example.com/login"
```

The login signs in at the instance's `/login` (password → `session_token` cookie), then spends that session once to mint a **never-expiring** personal API token (`POST <origin>/api/tokens`) and stores it as an `Authorization: Bearer` credential — durable, unlike the session cookie itself.

## Injects into the instance host *and its app subdomains*

The point of an OpenHost owner token is to sign in to the **client apps** the instance hosts — served on subdomains of the instance host (`<app>.<host>`). The router gates every non-public app path with the same owner auth that accepts this bearer, so the token is exactly the credential for reaching them (verified live: with it injected, protected app subdomains go 302 → 200; the router strips the session cookie and `X-OpenHost-*` before forwarding, so identity can't be spoofed to the backend).

A registered service previously matched only its single base prefix. Added `Service.registeredBaseApiUrls(baseApiUrl)`, which a family uses to decide what a self-hosted instance matches (default: the base itself); `RegisteredService` takes its match patterns from the family. OpenHost widens it to the instance host and any subdomain of it, anchored so a lookalike (`<host>.evil.com`) doesn't match.

## Base URL is the instance root, mint path is absolute

Register the instance **root** as the base API URL, not `.../api/`. OpenHost's owner API is not confined to `/api/`: app controls like `/stop_app/<id>`, `/reload_app/<id>`, `/restart_router`, `/toggle-ssh`, and `/app_logs/<id>` are served at the root, and the official `oh` CLI stores the base as `https://<host>` and calls absolute paths. A `/api/`-scoped base would leave the token uninjected on all those. The mint endpoint is a fixed absolute path (`/api/tokens`) on the instance origin, independent of that base.

## Family + browser-login binding, and a real credential status

Two `RegisteredService` gaps that only a family exercises, both fixed here:

- **Login binding.** A family-registered instance has to sign in at *its own* URL. It didn't — `RegisteredService` bound the family's session to the family, so the session read the family's URLs. No built-in service used the family+browser-login path, so this was latent. Fixed by binding the family's `getSession` to the registered service.
- **Credential status.** A registered service always reported status "unknown". Added `Service.registeredCredentialCheckCurlArguments(baseApiUrl)` (default: none → unknown, unchanged for generic registrations); OpenHost returns the instance's owner-guarded `/dashboard` (200 for a valid credential, a redirect otherwise — the official `oh` CLI validates the same way), so `services info` reports valid/invalid.

## Testing

- Unit tests (`tests/openhost.test.ts`): the mint call (subprocess runner stubbed), cookie detection, the instance-aware login binding, the host+subdomain match pattern (incl. lookalike rejection), and the credential status (valid on 200, invalid on the redirect, unknown for a generic registration) — through a real `RegisteredService` registration.
- Typecheck / lint / format clean.
- Proven live against a real self-hosted instance: register with the root → browser login mints the token → `latchkey curl` injects it for the owner API (`/api/apps`, `/dashboard`, `/app_logs/<id>`) and for protected app subdomains (`<app>.<host>`) alike; `services info` reports the credential valid, and a bogus token invalid.

## Note

Supersedes the earlier `token-mint` generic login-flow commit on this branch — that logic only existed for OpenHost, so it now lives in `services/openhost.ts` rather than as a core login-flow primitive.


boweiliu:
Superseded by #120 — same commits, branch renamed to `add-openhost-service-connector` (GitHub closed this PR when its head branch was renamed).

PR #122, opened 2026-08-13
Add DocuSign session-riding connector

boweiliu:
Adds a DocuSign eSignature connector.

## Approach
DocuSign's REST API authenticates with a short-lived (8h) bearer, not cookies, and its web SPA is issued no refresh token -- so neither the cookie-injection pattern (Slack) nor the OAuth-refresh pattern (dropbox/ramp/zoom) fits. This connector rides the user's web session: browser login captures the session cookies (+ initial bearer), and `refreshCredentials` re-mints the 8h bearer by loading the app with the stored cookies and capturing the `Authorization: Bearer` off the SPA's own API calls -- no password after the first login. Cookies last ~30 days.

## Refresh: just-in-time, stealth-browser-if-present
The mint needs a browser DocuSign does not block. Plain Chrome is blocked; a non-Chrome Chromium (Brave, or a stealth build) is not. latchkey stays BYO-browser -- it does not bundle one -- so `refreshCredentials` mints only when a refresh browser is configured via `LATCHKEY_REFRESH_BROWSER_PATH`; otherwise it returns null and latchkey falls back to interactive re-login (the ~8h path). With a refresh browser reachable, the session stays live ~30 days with no scheduler and no external process (refresh is driven by latchkey's existing expired-on-use hook).

## Contents
- `src/services/docusign.ts`: the connector + a custom `DocusignSessionCredentials` (cookies + bearer + expiry).
- Registered in `src/services/index.ts`, `src/serviceRegistry.ts`, and the credential serialization union.
- Unit tests for the credential (inject, expiry window, serialization roundtrip).

End-to-end verified against a live account: sending a contract for signature (POST envelope) returns 201 using only the stored session.

---
## Part of a 4-change set (DocuSign connector for latchkey/Minds)
Each is reviewable on its own, but they ship in dependency order:

1. **detent#27** -- the `docusign-api` enforcement schema (which hosts, read vs write). Foundation the others build on.
2. **latchkey#122** -- the DocuSign session-riding connector; uses the scope/permission names defined in #1.
3. **mngr-internal#371** (draft) -- the permission-catalog entry that makes the scope requestable in Minds; depends on #1 and #2 being released, then bumping this repo to those releases.
4. **(follow-up, Minds desktop -- being scoped)** -- wires the connector's token refresh to a non-Chrome browser so the connection lasts ~30 days instead of re-prompting every ~8h. Optional: without it, everything still works on the ~8h re-login path.

### DIFF
--- a/src/services/docusign.ts
+++ b/src/services/docusign.ts
@@ -160,7 +160,11 @@ export class DocusignSessionCredentials implements ApiCredentials {
   }
 
   static fromJSON(data: DocusignSessionData): DocusignSessionCredentials {
-    return new DocusignSessionCredentials(data.cookies, data.accessToken, data.accessTokenExpiresAt);
+    return new DocusignSessionCredentials(
+      data.cookies,
+      data.accessToken,
+      data.accessTokenExpiresAt
+    );
   }
 }
 
@@ -333,7 +337,9 @@ export class Docusign extends Service {
    * null so latchkey falls back to an interactive re-login. Returns null (never throws):
    * a thrown error here breaks callers like `services info` and the approval prompt.
    */
-  override async refreshCredentials(apiCredentials: ApiCredentials): Promise<ApiCredentials | null> {
+  override async refreshCredentials(
+    apiCredentials: ApiCredentials
+  ): Promise<ApiCredentials | null> {
     if (!(apiCredentials instanceof DocusignSessionCredentials)) {
       return null;
     }
--- a/src/services/docusign.ts
+++ b/src/services/docusign.ts
@@ -1,0 +1,369 @@
+/**
+ * DocuSign connector (session-riding, with a just-in-time bearer mint).
+ *
+ * Why this shape: DocuSign's eSignature REST API authenticates with a short-lived (8h)
+ * BEARER token, not cookies -- so we cannot inject cookies directly the way the Slack
+ * connector does. And its web SPA is issued NO refresh token, so we cannot refresh over
+ * plain HTTP the way the OAuth connectors (dropbox/ramp/zoom) do.
+ *
+ * What we CAN do (verified live): the user's session COOKIES silently authenticate the
+ * SPA, which then makes its own authenticated API calls. So the stored credential is the
+ * session cookies (like Slack's `d` cookie), plus the most recently seen bearer. To get a
+ * fresh bearer we do NOT try to reproduce DocuSign's token endpoint (that needs DocuSign's
+ * confidential web-client secret); instead we load the app with the cookies and CAPTURE
+ * the `Authorization: Bearer` header off the SPA's own requests. No password after the
+ * first login. Cookies last ~30 days, so the user re-auths about monthly.
+ *
+ * Refresh is just-in-time and needs a browser DocuSign does not block. Plain Chrome is
+ * blocked (headless and headed); a non-Chrome Chromium (Brave, or a stealth build like
+ * Fortress) is not. latchkey does not bundle such a browser -- it stays BYO-browser -- so
+ * `refreshCredentials` mints only when a refresh browser is configured
+ * (`LATCHKEY_REFRESH_BROWSER_PATH`); otherwise it returns null and latchkey falls back to
+ * an interactive re-login (the ~8h path). Point that variable at a Brave/Fortress-class
+ * executable to get the ~30-day silent path.
+ *
+ * Dynamic host: eSignature calls go to `{base_uri}/restapi/v2.1/accounts/{account_id}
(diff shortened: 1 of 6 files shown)


src/services/docusign.ts
```diff
@@ -0,0 +1,369 @@
...
+  readonly baseApiUrls = [DOCUSIGN_BASE_API_URL_PATTERN] as const;
+  readonly loginUrl = DOCUSIGN_LOGIN_URL;
+  readonly info =
+    'DocuSign eSignature REST API. After login, GET ' +
+    'https://account.docusign.com/oauth/userinfo to read your account_id and base_uri ' +
+    '(the regional host, e.g. https://na4.docusign.net); then call ' +
+    '{base_uri}/restapi/v2.1/accounts/{account_id}/... To send a contract for signature, ' +
+    'POST an envelope with status "sent". ' +
+    'Docs: https://developers.docusign.com/docs/esign-rest-api/. ' +
+    '(Auth rides the user\'s web session; the gateway re-mints the 8h bearer from the ' +
```
boweiliu: lets explicitly add some user/agent-facing guidance to avoid the docusign developer / oauth flow and docs since they're locked behind payment and procedure

src/services/docusign.ts
```diff
@@ -0,0 +1,369 @@
+/**
+ * DocuSign connector (session-riding, with a just-in-time bearer mint).
+ *
```
boweiliu, line 7: comments are a bit excessive, see working-sessions commentary from last time

src/services/docusign.ts
boweiliu: Done in cc8ab95 — added explicit `info` guidance to NOT set up a DocuSign integration key / OAuth app / go-live (that path is gated behind payment + review), and to just call the REST API via the browser session instead. Also dropped the developers.docusign.com docs link.

src/services/docusign.ts
boweiliu: Trimmed the file-header comment down to the essentials (one statement per idea) in cc8ab95.

hynek-urban:
@boweiliu Thanks for the contribution! People have already started asking for this one :)

I'm just not confident in the way how credentials are being refreshed here:
- there's a new undocumented env var which (needlessly?) increases the control surface
- by default, non-technical users would struggle (imagine this being part of the Minds app as is)

Is there a reason to not just using the normal headless browser?

In fact, let me push back even more. it would be significantly better to not use the browser for the refresh logic at all. (Browsers are heavy and spinning them up in the background for an operation that's meant to be as lightweight as refreshing credentials goes against the [Principle of least asthonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Surely the way how the browser exchanges the cookie for the bearer token is fundamentally reducible to a series of requests. Is there any chance we could reverse-engineer the necessary request sequence and replay it instead of spinning up the whole browser?

boweiliu:
topic 1 - yeah pulling up evidence and also looking for other workarounds.
topic 2 in the other PR - lgtm will do the path filter.

boweiliu:
also we should fix merge conflicts

boweiliu:
@hynek-urban Good push — I dug into exactly this. Short version: **step 1 reduces to a request; step 2 fundamentally needs a JS engine.**

The mint is two steps:
1. `GET account.docusign.com/oauth/auth?...&client_id=e63ccab8&redirect_uri=apps.docusign.com/authenticate` + session cookie -> a fresh `code`, silently. **Pure HTTP, replayable.**
2. `apps.docusign.com/authenticate?code=...` -> returns the SPA HTML; **the SPA's JavaScript** turns the code into the token in `localStorage["@1ds/shell.auth_info"]`.

I replayed the whole silent chain with a pure HTTP client (cookie jar + redirects, no page/JS): final **HTTP 200, body = 74,685 bytes of HTML+JS, and no token**. The token is never returned by an endpoint — it is materialized client-side by DocuSign's own JS (an earlier browser-XHR scan also found 0 responses containing `access_token`). So "emulate it with JS" means running DocuSign's proprietary web-auth JS, which rides the **confidential** web client (PKCE-only exchange -> `unauthorized_client`) — brittle, and no lighter than a headless browser.

On the browser choice: DocuSign's anti-bot blocks plain Chrome (headless *and* headed), so the mint needs a non-Chrome Chromium (Brave / a stealth build) — that's the only reason for `LATCHKEY_REFRESH_BROWSER_PATH`. To keep it light the connector launches headless **only at refresh, just-in-time, only when such a browser is configured**, and otherwise returns null and degrades to an ~8h interactive re-login.

Fair point on the control surface — a config field would be cleaner than an env var, and in Minds we'd point it at a browser the gateway already has so a user configures nothing. Full write-up with the repro script: working-sessions/250 in my notes repo. Happy to walk through it.

PR #123, opened 2026-08-13
Add Tailscale connector

boweiliu:
Adds a Tailscale service connector for the admin API at `api.tailscale.com/api/v2`.

## Approach

A browser login (`latchkey auth browser tailscale`) signs the user into the admin console and mints an **API access token** (`tskey-api-...`) from the Keys page's one-time reveal dialog, capturing the tailnet name from the navigation. The token carries the full set of actions its owning user may perform — tailnet-wide user management (role changes, approval, suspension, deletion, restoration), device management, ACL/policy, DNS, and the minting of further keys including the auth keys the `tailscale` CLI consumes — and expires after at most 90 days, after which the user re-runs the login. Mirrors the OpenRouter/HuggingFace/Linear `BrowserFollowupServiceSession` mint-and-reveal pattern.

## Why the API access token (not an OAuth client)

Tailscale also offers OAuth clients (long-lived client_id/secret + a 1-hour client-credentials flow, no 90-day re-login), which are operationally cleaner for a long-running automation. They are a worse fit for the *autonomous* connector, though: their mint dialog requires choosing scopes and (for `auth_keys` write) org-specific tags, so a followup session cannot pick them without a mid-flow human decision. The API access token's mint is a simple description + expiry + one-time reveal, and a single token serves every admin action an Owner/Admin can take — including minting CLI auth keys with no tag constraint. An OAuth-client connector (`tailscale-oauth`) is a reasonable follow-up for a read-only, no-re-login workload.

## Credential shape

`TailscaleCredentials` (token + tailnet) injects an `Authorization: Bearer` header into every request to `api.tailscale.com`. The tailnet is part of nearly every useful API path and the API exposes no tailnet-independent way to validate a token, so it is captured at mint time and stored alongside the token. The credential check hits `GET /tailnet/{tailnet}/settings` (200 → valid, 401/404 → invalid); a manually-set credential (no tailnet) reports `unknown` while still injecting normally for `latchkey curl`. `getAccount` returns the tailnet.

## Contents

- `src/services/tailscale.ts` — the connector + `TailscaleCredentials`.
- Registered in `src/services/index.ts` and `src/serviceRegistry.ts`; `TailscaleCredentials` added to the credential serialization union in `src/apiCredentials/serialization.ts`.
- Listed in both `SKILL.md` service lists; blacklisted in `tests/servicesAgainstRecordings.test.ts` (a `BrowserFollowupServiceSession` whose token is minted by a one-time reveal dialog, like linear/ngrok/huggingface/openrouter).
- `tests/tailscale.test.ts` — URL matching, credential injection, account, serialization round-trip, and the credential check (valid / 401 / 404 / non-tailscale → unknown).

## Verified live

Against a real tailnet (imbue.com): a browser login minted a sealed `tskey-api` token; the connector's credential check reported it `valid`; `latchkey curl`-equivalent injection reached `GET /tailnet/{tailnet}/users`; a tampered tailnet and a revoked token both reported `invalid`. The token never left the gateway; all test credentials were revoked.

lint, typecheck, and `npm test` (766 tests) pass; the 6 CLI subprocess-suite flakes fail identically on clean `main` (sandbox gateway environment).

Pairs with a Detent scopes PR (`tailscale-api` + read/write) and an mngr-internal catalog entry, to follow.

### DIFF
--- a/src/services/tailscale.ts
+++ b/src/services/tailscale.ts
@@ -1,0 +1,273 @@
+/**
+ * Tailscale service implementation.
+ *
+ * The Tailscale admin API (https://api.tailscale.com/api/v2/) authenticates with
+ * an API access token (`tskey-api-...`) that an Owner/Admin creates in the
+ * admin console's Keys page. The token's value is revealed exactly once at
+ * creation time (like Linear, HuggingFace and OpenRouter), so we sign the user
+ * into the admin console, generate a fresh access token on their behalf, and
+ * read its value from the one-time reveal dialog.
+ *
+ * A single API access token grants the full set of actions its owning user may
+ * perform: tailnet-wide user management (role changes, approval, suspension,
+ * deletion), device management, ACL/policy, DNS, and the minting of further
+ * keys (including the auth keys the `tailscale` CLI consumes to bring up
+ * nodes). It expires after at most 90 days, after which the user re-runs the
+ * browser login to mint a new one -- the token is static and is not refreshable.
+ */
+
+import type { Response, BrowserContext } from 'playwright';
+import { z } from 'zod';
+import { ApiCredentialStatus, ApiCredentials } from '../apiCredentials/base.js';
+import { typeLikeHuman } from '../playwrightUtils.js';
+import { runCapturedAsync } from '../curl.js';
+import {
+  Service,
+  BrowserFollowupServiceSession,
+  FollowupWork,
+  LoginFailedError,
+  isBrowserClosedError,
+  LoginCancelledError,
+} from './core/base.js';
+
+const DEFAULT_TIMEOUT_MS = 12000;
+
+// The Tailscale admin REST API.
+const TAILSCALE_API_BASE_URL = 'https://api.tailscale.com/';
+
+// `latchkey auth browser` opens this; it redirects to the SSO sign-in page and
+// then back to the admin console once the user is signed in.
+const TAILSCALE_LOGIN_URL = 'https://login.tailscale.com/admin';
+
+// The admin console's Keys page, where API access tokens are generated.
+const TAILSCALE_KEYS_URL = 'https://console.tailscale.com/admin/settings/keys';
+
+// The admin console host; a document response served from it means the user is
+// signed in (the sign-in flow itself is served from login.tailscale.com and the
+// SSO provider).
+const TAILSCALE_CONSOLE_HOST = 'console.tailscale.com';
+
+// GET /tailnet/{tailnet}/settings is a cheap read that answers 200 for any
+// valid token with the access the API access token carries (full user powers).
+// It doubles as the credential check once the tailnet is known.
+const TAILSCALE_SETTINGS_URL_PREFIX = 'https://api.tailscale.com/api/v2/tailnet/';
+
+// API access tokens are prefixed `tskey-api-`; the reveal dialog renders the
+// value as plain text (not inside an input).
+const TAILSCALE_TOKEN_PATTERN = /tskey-api-[A-Za-z0-9-]+/;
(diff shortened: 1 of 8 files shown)


boweiliu:
## Recordings-blacklist paper trail

`tailscale` is added to the `BLACKLIST` in `tests/servicesAgainstRecordings.test.ts` because it is a `BrowserFollowupServiceSession`: its API access token is minted by an interactive browser action (the Keys page's one-time reveal dialog) and is therefore not present in the recorded login network traffic, so replaying a recording cannot reconstruct it. This is the same criterion the other browser-followup connectors are blacklisted under — `dropbox`, `github`, `linear`, `ngrok`, `huggingface`, `openrouter` — exactly like `linear` (its token is created and read from a one-time reveal dialog).

The blacklist entry also documents the known `todoist` gap (it predates this list and is left as-is to keep each change scoped), consistent with the note the ngrok connector added in #113.

boweiliu:
## Gateway end-to-end verification

Ran the connector through a live, isolated latchkey gateway (separate throwaway directory on `localhost:19890`, not the workspace gateway, to avoid hot-mod residue). The tailscale detent scopes were provided inline in `permissions.json` as user schemas; a fresh `tskey-api` token was minted from the dashboard and stored via `latchkey auth set tailscale`.

| test | request | grant | result |
|---|---|---|---|
| read allowed | `GET /api/v2/tailnet/imbue.com/users` | `tailscale-api: [tailscale-read-all]` | **200** — full user list returned |
| write denied | `POST /api/v2/tailnet/imbue.com/keys` | `tailscale-api: [tailscale-read-all]` | **403** — "Request not permitted by the user" |
| write allowed | `POST /api/v2/tailnet/imbue.com/keys` | `tailscale-api: [tailscale-read-all, tailscale-write-all]` | **200** — minted a real `tskey-auth` key |

This confirms the full path: the tailscale service's URL matching, credential injection, and detent scope enforcement all work end-to-end through the gateway. The read-only grant blocks writes (403) and the write grant allows them (200), so the scope boundary is real.

The browser-mint flow (`latchkey auth browser tailscale`) was verified separately earlier (the token minted by the dashboard reveal dialog drove `checkApiCredentials` -> `valid`, and a revoked token -> `invalid`). The full `auth browser` flow couldn't run inside this container (no GUI for the user's Google SSO), so the token was minted via the dashboard for the gateway test instead.

All test credentials were revoked (the `tskey-auth` via `DELETE /keys`, the `tskey-api` via the dashboard) and confirmed dead (401). The isolated gateway was torn down and the throwaway directory removed — no residue.

src/services/tailscale.ts
```diff
@@ -0,0 +1,273 @@
...
+  readonly baseApiUrls = [TAILSCALE_API_BASE_URL] as const;
+  readonly loginUrl = TAILSCALE_LOGIN_URL;
+  readonly info =
+    'Tailscale admin API for tailnet-wide management (users, devices, policy, keys). ' +
+    'Docs: https://tailscale.com/docs/reference/tailscale-api and interactive ' +
+    'https://tailscale.com/api. A browser login mints an API access token (tskey-api) ' +
+    'that carries the full set of actions its owning user may perform and expires ' +
+    'after at most 90 days; re-run the login to mint a new one. Most endpoints are ' +
+    'scoped to a tailnet (e.g. /api/v2/tailnet/{tailnet}/users); the stored credential ' +
+    'remembers the tailnet seen at login.';
```
boweiliu, line 219: unsure, please inform me - do we need all this info so verbosely? also do we need to mention tailscale api docs or cli flows?

hynek-urban:
LGTM!

PR #124, opened 2026-08-14
google-gmail: request gmail.settings.basic scope for filter management

boweiliu:
## Problem

The Gmail service requested only `https://www.googleapis.com/auth/gmail.modify` at login. That scope permits reading/labeling messages and reading filters/labels, but **creating or deleting filters** (`users.settings.filters`) requires `https://www.googleapis.com/auth/gmail.settings.basic`.

As a result, any agent or client attempting to manage Gmail filters via `users.settings.filters.*` failed with:

```
403 ACCESS_TOKEN_SCOPE_INSUFFICIENT
MailboxService.CreateFilter
```

…even after a fresh login. The existing devNote ("logging in again and approving all the scopes might help") was misleading here, because the scope was never requested — so re-login changed nothing.

## Fix

Add `gmail.settings.basic` to the gmail service's requested scopes, so newly minted tokens can manage filters. Extend the devNote to call out that older tokens minted without this scope must be re-logged-in to gain it.

This is additive: `gmail.modify` is retained (it covers message/label read+write and filter *read*); `gmail.settings.basic` adds filter create/delete.

## Migration note

Existing stored tokens do not gain the new scope automatically. Users who want to manage filters must re-login (`latchkey` will request the expanded scope set, and the user must approve it in the Google consent screen). The devNote now states this.

## Test

Adds `tests/googleGmailScopes.test.ts`, a scope-regression test asserting the gmail service config contains both `gmail.modify` and `gmail.settings.basic`. Prevents an accidental scope drop that would silently re-break filter management.

`npm run typecheck` ✅ · `npm run lint` ✅ · `npx vitest run tests/googleGmailScopes.test.ts tests/serviceRegistry.test.ts` ✅ (77/77).

---

Discovered while building Gmail filter automation: creating filters for `bowei@imbue.com` failed with `ACCESS_TOKEN_SCOPE_INSUFFICIENT`; root-caused to the single-scope `CONFIG.scopes` in `src/services/google/gmail.ts`.


src/services/google/gmail.ts
```diff
@@ -14,7 +19,8 @@ export class GoogleGmail extends GoogleService {
     'If needed, run "latchkey auth browser-prepare google-gmail" to create an OAuth client first. ' +
     'It may take a few minutes before the OAuth client is ready to use. ' +
     'Requests that end with `ACCESS_TOKEN_SCOPE_INSUFFICIENT` may be caused by some scopes not having been approved during login. ' +
-    'Logging in again and approving all the scopes might help in that case.';
+    'Logging in again and approving all the scopes might help in that case. ' +
+    'Note: filter create/delete requires the `gmail.settings.basic` scope, which is now requested at login; older tokens minted without it must be re-logged-in to gain it.';
```
boweiliu: hey this comment doesnt seem helpful going forward esp after we merge this. maybe move it to PR description?

boweiliu: Good catch — removed that sentence from the `info` string in 4eb3800. The migration note now lives only in the PR description. The one-line code comment above the scopes (explaining what each scope covers) stays since it's useful for future maintainers.

PR #125, opened 2026-08-14
Add Spotify connector (web-player session + lazy browser re-mint)

boweiliu:
## What

Adds Spotify as a first-class latchkey service with a custom `SpotifySessionCredentials` credential type and a lazy browser-based re-mint.

## Why

Spotify has no personal API key and the official OAuth Developer-App path requires Spotify Premium (blocked for Free accounts as of Feb 2026). This connector rides the web player's own session: a one-time browser login captures the long-lived `sp_dc` cookie (~1 year, HttpOnly), and a short-lived bearer (~10-15 min) is minted from it on demand by running a headless browser through the web player's own `/api/token` flow (the TOTP is computed by Spotify's JS, so a browser must run the mint).

The connector targets `spclient.wg.spotify.com` (the web player's own REST-like API, 200-confirmed) rather than `api.spotify.com/v1` (which rate-limits the web-player token).

## How

- **Acquisition:** `BrowserFollowupServiceSession` — user logs into Spotify in a headful Chrome; the followup reads `sp_dc` + captures the initial bearer from the SPA's `/api/token` response.
- **Injection:** `Authorization: Bearer <accessToken>` on every spclient call.
- **Refresh:** lazy mint — `refreshCredentials` launches a headless Chrome with `sp_dc` set, loads `open.spotify.com` so the SPA computes the TOTP and hits `/api/token`, captures the response bearer + `accessTokenExpirationTimestampMs`, returns new credentials. Mint happens only when a call arrives with an expired token (no idle minting).
- **Manual set:** `auth set-nocurl spotify <sp_dc-value>` stores `sp_dc` only; the bearer is minted on first use.

New credential type: `SpotifySessionCredentials` (`objectType: spotifySession`), registered in the Zod discriminated union + serialize/deserialize in `serialization.ts`.

### Login-complete detection (two fixes)

Spotify's login flow has two quirks that required fixes beyond the standard `BrowserFollowupServiceSession` pattern:

1. **SPA route changes don't fire document responses.** After the user completes the email-code login, the browser lands on `accounts.spotify.com/en/status` — but this happens as a client-side SPA navigation, not a new document response. The standard `onResponse` handler (which only checks `resourceType === 'document'`) never fires for `/en/status`, so the connector hangs waiting for login-complete. **Fix:** `isLoginComplete` polls `page.url()` directly (every 100ms via the base class's `waitForLoginComplete` loop), not just the `onResponse` document flag. The page reference is captured in `onResponse` via `response.request().frame().page()`.

2. **The mint only fires on `open.spotify.com`.** When login lands on `accounts.spotify.com/en/status`, the `/api/token` mint doesn't fire there — it only fires when the web player SPA (`open.spotify.com`) loads. The followup was listening for `/api/token` on the `/en/status` page, which never fires it. **Fix:** `performBrowserFollowup` checks `page.url()` and navigates to `open.spotify.com` if the page isn't already there, before starting the mint listener.

## Tested

E2E through a live Minds gateway using the full latchkey permission-approval + browser-login flow (no manual credential tricks):
- `latchkey auth browser spotify` opens Chrome → user logs in (captcha + password + email code) → connector detects login-complete on `accounts.spotify.com/en/status` → navigates to `open.spotify.com` → SPA fires `/api/token` → bearer + `sp_dc` captured → credential stored.
- `latchkey curl https://spclient.wg.spotify.com/user-profile-view/v3/profile/me` → 200 with real profile data.
- Playlist CRUD (create + delete) confirmed via `latchkey curl`.

## Pairs with

- `imbue-ai/detent#29` — the `spotify-api` Detent scope schema.
- `imbue-ai/mngr-internal#396` — the `spotify-api` catalog entry in `services.json`.

### DIFF
--- a/src/services/spotify.ts
+++ b/src/services/spotify.ts
@@ -1,0 +1,454 @@
+/**
+ * Spotify service implementation.
+ *
+ * Spotify is a *session-cookie + silent re-mint* service (the "Shape 6"
+ * connector from the latchkey-dev-workbench design doc). The web player holds a
+ * long-lived `sp_dc` cookie (~1 year) and silently mints a short-lived bearer
+ * (~10-15 min) via `https://open.spotify.com/api/token?...&totp=<TOTP>`. The
+ * TOTP is computed by the SPA's own JS, so the mint is NOT HTTP-reducible; a
+ * browser must run it.
+ *
+ * There is no personal API key (so no key scrape), and the official Developer
+ * App (OAuth) path requires Spotify Premium (blocked for Free accounts as of
+ * Feb 2026). The public `api.spotify.com/v1` is rate-limited (429) for the
+ * web-player token; the connector therefore rides the web player's own API at
+ * `spclient.wg.spotify.com` (REST-like, confirmed 200 with this token).
+ *
+ * Auth shape:
+ * - Acquisition: a browser login (modelled on ngrok's BrowserFollowupServiceSession).
+ *   The user signs into accounts.spotify.com → open.spotify.com; once logged
+ *   in, the followup step reads `sp_dc` from the context's cookies and mints
+ *   the initial bearer by capturing the SPA's own `/api/token` response.
+ * - Injection: `Authorization: Bearer <accessToken>` on every spclient call.
+ * - Refresh: `refreshCredentials` does a LAZY headless re-mint — launches a
+ *   headless Chrome with `sp_dc` set, loads open.spotify.com so the SPA
+ *   computes the TOTP and hits /api/token, captures the response bearer + its
+ *   `accessTokenExpirationTimestampMs`, returns new credentials. Mint happens
+ *   ONLY when a call arrives with an expired token (no idle minting).
+ *
+ * This rides Spotify's undocumented first-party web client (ToS-gray) and can
+ * break if Spotify changes the web login/mint. The durable path is an upstream
+ * latchkey PR + the sanctioned OAuth connector if the account ever goes Premium.
+ */
+import type { BrowserContext, Page, Response } from 'playwright';
+import { z } from 'zod';
+import {
+  ApiCredentials,
+  ApiCredentialsUsageError,
+  ApiCredentialStatus,
+} from '../apiCredentials/base.js';
+import { loadPlaywright } from '../playwrightLoader.js';
+import {
+  Service,
+  BrowserFollowupServiceSession,
+  FollowupWork,
+  LoginFailedError,
+} from './core/base.js';
+
+const SPOTIFY_LOGIN_URL = 'https://accounts.spotify.com/en/login';
+const SPOTIFY_WEB_PLAYER_URL = 'https://open.spotify.com/';
+const SPOTIFY_API_TOKEN_URL_PREFIX = 'https://open.spotify.com/api/token';
+const SPOTIFY_OPEN_HOST = 'open.spotify.com';
+const SPOTIFY_ACCOUNTS_HOST = 'accounts.spotify.com';
+// The /en/status path on accounts.spotify.com is the post-login success landing
+// (the browser may stay there instead of redirecting to open.spotify.com).
+const SPOTIFY_PRE_LOGIN_PATH_PATTERN = /^\/(login|signup|signin|auth|sso|oauth|verify|mfa|otp)/i;
+
+// The web player's own API (REST-like). Confirmed 200 with the web-player bearer.
+// The web player's own API (REST-like). Confirmed 200 with the web-player bearer.
+// All the data-API calls (profile, playlists, library) go to this single host;
+// the guc3-spclient.spotify.com variant is for playback-state, not the data API.
+const SPOTIFY_API_BASE_URLS = ['https://spclient.wg.spotify.com/'] as const;
+
+// --- Credentials ---
+
+/**
+ * Stored Spotify credentials: the durable `sp_dc` session cookie, plus a cached
+ * short-lived `accessToken` and its expiry (ms epoch). The access token is
+ * minted from `sp_dc` on demand (see {@link SpotifyService.refreshCredentials});
+ * there is no refresh token.
+ */
+export const SpotifySessionCredentialsSchema = z.object({
+  objectType: z.literal('spotifySession'),
+  sp_dc: z.string(),
+  accessToken: z.string().optional(),
+  accessTokenExpiresAt: z.number().optional(), // ms epoch
+});
+export type SpotifySessionCredentialsData = z.infer<typeof SpotifySessionCredentialsSchema>;
+
+export class SpotifySessionCredentials implements ApiCredentials {
+  readonly objectType = 'spotifySession' as const;
+  constructor(
+    readonly sp_dc: string,
+    readonly accessToken?: string,
+    readonly accessTokenExpiresAt?: number
+  ) {}
+
+  injectIntoCurlCall(curlArguments: readonly string[]): Promise<string[]> {
+    if (!this.accessToken) {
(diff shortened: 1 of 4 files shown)


src/services/spotify.ts
```diff
@@ -0,0 +1,423 @@
...
+
+export class Spotify extends Service {
+  readonly name = 'spotify';
+  readonly displayName = 'Spotify';
+  readonly baseApiUrls = SPOTIFY_API_BASE_URLS;
+  readonly loginUrl = SPOTIFY_LOGIN_URL;
+  readonly info = [
+    'Spotify (web-player session connector).',
+    '',
+    "Rides the Spotify web player's own session: a one-time browser login",
```
boweiliu, line 311: is this info strictly all necessary? (i am a bit confused what it means and what's the intended audience fwiw). 

also do we need to call out more specific instructions to prefer the "partner" spclient api, and where to find its docs?

src/services/spotify.ts
```diff
@@ -0,0 +1,423 @@
...
+        'Spotify refresh expected SpotifySessionCredentials',
+      );
+    }
+    const sp_dc = apiCredentials.sp_dc;
+    if (!sp_dc) {
+      return null; // nothing to mint from -> user must re-login
+    }
+
+    const playwright = await loadPlaywright();
+    // Use the system Chrome if available (the bundled Playwright Chromium's
```
boweiliu, line 400: are you sure this is true? seems sus, both the code logic and the description.

src/services/spotify.ts
boweiliu: You're right — the info string was too verbose and mixed implementation detail with user-facing notes. I've trimmed it in the latest push to the essentials: what the connector does (rides the web player session), the target API (spclient.wg.spotify.com), and why not the public v1 API (429).

On the spclient docs question: there are no public docs — it's Spotify's internal web-player API, undocumented. The surface is discoverable via the browser network tab (REST-like routes: /user-profile-view/v3/profile/{user}, /playlist/v2/playlist, /playlist/v2/user/{user}/rootlist/changes, etc.). I've noted in the info string that it's undocumented and discoverable via the browser. I'd lean against baking a route list into the code since Spotify can change these at any time, but happy to add one to the PR description if useful.

src/services/spotify.ts
boweiliu: Fair point — the comment was misleading. The actual situation: the system Chrome (Google Chrome.app) is used as the executablePath because the bundled Playwright Chromium's headless-shell binary (`chrome-headless-shell`) isn't installed by default on the user's Mac — only the full Chromium is. When `headless: true` is used, Playwright looks for the headless-shell variant first, which doesn't exist, causing a crash. The system Chrome supports `--headless` natively so it works as a drop-in.

I've updated the comment to be more precise: the issue is the headless-shell variant not being installed, not that the system Chrome is "always present" (it's macOS-specific; on Linux the bundled chromium would work fine). The `executablePath` falls back to `undefined` (bundled chromium) if the system Chrome isn't found.

If you'd prefer a different approach (e.g. running `npx playwright install` as a prerequisite, or using `headless: false` with a virtual framebuffer), let me know — but the system-Chrome fallback is the simplest path for the Mac-host case.

research/tailscale-openapi.json
```diff
@@ -0,0 +1,6683 @@
+openapi: 3.1.0
```
boweiliu: why did these tailscale things get included? seems irrelevant

src/services/spotify.ts
```diff
@@ -0,0 +1,422 @@
...
+const SPOTIFY_API_TOKEN_URL_PREFIX = 'https://open.spotify.com/api/token';
+const SPOTIFY_OPEN_HOST = 'open.spotify.com';
+const SPOTIFY_PRE_LOGIN_PATH_PATTERN =
+  /^\/(login|signup|signin|auth|sso|oauth|verify|mfa|otp|status)/i;
+
+// The web player's own API (REST-like). Confirmed 200 with the web-player bearer.
+const SPOTIFY_API_BASE_URLS = [
+  /^https:\/\/[a-z0-9-]+\.spclient\.wg\.spotify\.com\//,
+  'https://spclient.wg.spotify.com/',
+  /^https:\/\/[a-z0-9-]+\.spclient\.spotify\.com\//,
```
boweiliu: are you sure we need the wildcard prefix? cite evidence

boweiliu: if you think we dont need them, also cite evidence

research/tailscale-openapi.json
boweiliu: You're right, those were accidentally included — they're from an unrelated research task in the cloned repo. Removed in the latest push; the diff is now clean (only the 4 spotify-related files).

src/services/spotify.ts
boweiliu: You're right to question this. Evidence from the CDP capture:

- **spclient.wg.spotify.com** — the data API (profile, playlists, library). All the 200-confirmed calls the connector targets go here.
- **guc3-spclient.spotify.com** — only used for playback-state (connect-state, track-playback), NOT the data API.

The wildcard `[a-z0-9-]+\.spclient\.spotify\.com` was meant to catch the `guc3-` prefix, but that's the playback-state host — the connector doesn't target it. I've narrowed `baseApiUrls` to just `'https://spclient.wg.spotify.com/'` (the single host the data API lives on). The Detent scope schema (imbue-ai/detent#29) is similarly narrowed to `domain: {const: spclient.wg.spotify.com}`.

research/tailscale-openapi.json
boweiliu: i still see them

boweiliu: Apologies — the earlier amend removed them from the index but not from the tree. Now fully removed: `git ls-tree` confirms no `research/` files in the branch, and the diff is clean (4 files, 430 insertions).

src/services/spotify.ts
```diff
@@ -0,0 +1,454 @@
...
+/**
+ * Stored Spotify credentials: the durable `sp_dc` session cookie, plus a cached
+ * short-lived `accessToken` and its expiry (ms epoch). The access token is
+ * minted from `sp_dc` on demand (see {@link SpotifyService.refreshCredentials});
+ * there is no refresh token.
+ */
+export const SpotifySessionCredentialsSchema = z.object({
+  objectType: z.literal('spotifySession'),
+  sp_dc: z.string(),
+  accessToken: z.string().optional(),
```
hynek-urban, line 74: Is this optional just to support `getCredentialsNoCurl()`? If so then I'd say we don't need to bother with it (I don't think anyone will manually insert the cookie value on the terminal?). What do you think?

hynek-urban (commented):
Thank you, @boweiliu :) This one is exciting.

Similar to the discussion [here](https://github.com/imbue-ai/latchkey/pull/122#issuecomment-5294551354), do you think there's any chance that we could actually run "Spotify's JS" to compute the ToTP? We can run JS in Latchkey :)

hynek-urban:
Actually, I just realized that this hits the undocumented private API :( I think we generally don't want to make undocumented APIs part of Latchkey as a policy, for several reasons:

a) Since the API is undocumented, agents don't really know how to use it and there's no documentation to point them to it.
b) In this case, not only is it against Spotify's ToS but it seems they actually actively ban accounts of people who use the internal API with a scraped token. I think that's a risk we shouldn't expose our users to.

hynek-urban:
@boweiliu This was turned into an example Latchkey plugin, see https://github.com/imbue-ai/latchkey/blob/main/docs/plugins.md :)

Issue #127, opened 2026-08-22
Allow a user-registered service to match multiple hosts (repeatable --base-api-url, or a pattern)


thad-imbue:
### Problem

A user-registered service can only ever match **one** URL prefix, so an API
that spans more than one host needs one registration per host — and one
`latchkey auth set` per registration, even when a single token covers all of
them.

Fastmail is a clean example. The JMAP API lives on `api.fastmail.com`, blobs
(`{downloadUrl}`) on `www.fastmailusercontent.com`, and accounts homed on a
regional shard use *different* hosts again — `phl.api.fastmail.com` and
`phl-www.fastmailusercontent.com`. One API token covers all four, but setup is:

```sh
latchkey services register fastmail             --base-api-url="https://api.fastmail.com/"
latchkey services register fastmail-content     --base-api-url="https://www.fastmailusercontent.com/"
latchkey services register fastmail-phl         --base-api-url="https://phl.api.fastmail.com/"
latchkey services register fastmail-content-phl --base-api-url="https://phl-www.fastmailusercontent.com/"

latchkey auth set fastmail             -H "Authorization: Bearer $(pbpaste)"
latchkey auth set fastmail-content     -H "Authorization: Bearer $(pbpaste)"
latchkey auth set fastmail-phl         -H "Authorization: Bearer $(pbpaste)"
latchkey auth set fastmail-content-phl -H "Authorization: Bearer $(pbpaste)"
```

Four services and four `auth set` calls for one credential — and all four have
to be re-run every time the token rotates.

### The engine already supports what's needed

`ServiceRegistry.matchesUrl` handles both strings and patterns
(`src/serviceRegistry.ts`):

```js
matchesUrl(service, url) {
  for (const baseApiUrl of service.baseApiUrls) {
    if (typeof baseApiUrl === 'string') { if (url.startsWith(baseApiUrl)) return true; }
    else { if (baseApiUrl.test(url)) return true; }
  }
  return false;
}
```

`Service.baseApiUrls` is typed `readonly (string | RegExp)[]`, and built-ins
already use both the array and the regex form — `aws` is
`/^https:\/\/[^/]*\.amazonaws\.com\//`, `mailchimp` is
`/^https:\/\/[^/]+\.api\.mailchimp\.com\//`.

Three places close that off for user-registered services:

1. `services register` takes `--base-api-url <url>`, not repeatable.
[...]

PR #129, opened 2026-08-24
fastmail: add JMAP and DAV services

### LINKED ISSUE #128: Add Fastmail as a built-in service (regional shard hosts need pattern URLs)
### Request

Add Fastmail as a built-in service. Its JMAP API is a good fit for latchkey —
the credential is a plain user-facing API token (no OAuth dance, no cookie
capture, no TLS-fingerprint wall), but it spans several hosts, which a
user-registered service can't express today (see #127).

### Hosts

| purpose | host |
|---|---|
| JMAP API + session discovery | `api.fastmail.com` |
| JMAP API, region-homed account | `<shard>.api.fastmail.com`, e.g. `phl.api.fastmail.com` |
| Blob / attachment downloads | `www.fastmailusercontent.com` |
| Blob downloads, region-homed | `<shard>-www.fastmailusercontent.com`, e.g. `phl-www.fastmailusercontent.com` |

Which shard you get is a property of the account, and it isn't visible until
`GET https://api.fastmail.com/.well-known/jmap` hands back an `apiUrl` and
`downloadUrl` pointing at it. So a user can't know up front which hosts they'll
need.

### Suggested shape

Same pattern as the existing `mailchimp` / `aws` services:

```ts
baseApiUrls = [
  /^https:\/\/([a-z0-9-]+\.)?api\.fastmail\.com\//,
  /^https:\/\/[a-z0-9-]+\.fastmailusercontent\.com\//,  // www. and <shard>-www.
];
```

- Credential: `Authorization: Bearer <token>`, from
  <https://app.fastmail.com/settings/security/tokens> (read-only mail scope is
  enough for JMAP `urn:ietf:params:jmap:mail`).
- `checkApiCredentials` could `GET https://api.fastmail.com/.well-known/jmap`
  and look for `primaryAccounts` — 401 means the token is bad.
- `getAccount` has a natural answer too: the session document's `username`
  field is the account's email address.

That would collapse the whole setup to one `latchkey auth set fastmail`,
regardless of which shard the account lives on.

### One gotcha worth baking into the docs

The obvious smoke test is misleading on a shard host:

```sh
latchkey curl -sSL https://phl.api.fastmail.com/.well-known/jmap
# => No Authorization header
```

That request 302s cross-host to `api.fastmail.com/jmap/session`, and curl drops
the `Authorization` header on cross-host redirects — so it reports a missing
header even when the credential is registered and working. Hitting a
non-redirecting path on the same host shows auth is actually fine (405, not
401):

```sh
latchkey curl -sSi https://phl.api.fastmail.com/jmap/api/    # HTTP/2 405
curl -sSI          https://phl.api.fastmail.com/jmap/api/    # HTTP/2 401
```

Observed on latchkey 3.5.0.

thad-imbue:
Adds Fastmail as a built-in service. Closes #128.

Fastmail runs a full OAuth 2.1 authorization server — dynamic client
registration, PKCE S256, refresh tokens, public clients — so `latchkey auth
browser fastmail` is enough and no hand-made API token has to be pasted in.

Two services, split by host, because they take different credentials:

| service | hosts | credential |
|---|---|---|
| `fastmail` | `*.api.fastmail.com`, `*.fastmailusercontent.com` | OAuth (browser login) — mail, contacts, calendars over JMAP |
| `fastmail-dav` | `*carddav.fastmail.com`, `*caldav.fastmail.com` | app password over Basic auth |

### Three things the obvious implementation gets wrong

Each of these was found the hard way, against the live server.

**The account lives on a regional shard.** Session discovery starts at
`api.fastmail.com`, but the `apiUrl` and `downloadUrl` it returns point at hosts
like `phl.api.fastmail.com` and `phl-www.fastmailusercontent.com` — and which
shard that is is only knowable *after* discovery. This is why the service uses
pattern `baseApiUrls`. Done as a self-registered service with literal prefixes
it takes four registrations and four `auth set` calls to carry one credential,
which is what motivated #128.

**Scopes are fixed at client registration.** The authorization request may only
ask for a subset of what the client registered for, so registering without
`scope` produces a client with none and every later authorization fails. The
scope names are the ones the authorization server's own metadata advertises;
the JMAP URNs in Fastmail's developer docs (`urn:ietf:params:jmap:mail`, …)
describe hand-registered partner clients, and dynamic registration rejects them:

```
invalid_client_metadata invalid scopes: urn:ietf:params:jmap:mail
```

**The authorization request must carry an RFC 8707 `resource`**, or Fastmail
refuses it with `invalid_target`. Fastmail publishes two protected resources
with disjoint scope vocabularies — the JMAP session and an MCP endpoint — so the
token is audience-bound to whichever is named. The same `resource` has to ride
on the token exchange and on refreshes.

### Why split by host rather than by capability

`google-gmail` / `google-people` suggest a service per capability, but that
doesn't work here. `latchkey curl` picks among services matching a URL by
registration order and gives callers no way to name one, and JMAP multiplexes
mail, contacts and calendars through a single endpoint — `POST /jmap/api/`,
differing only in the request body's `using` array. Per-capability services
would all claim the same URL, and a request could go out holding a credential
without the scope it needs.

Host-based splitting has no such overlap, and it lines up with the credential
boundary anyway: OAuth tokens are accepted only by the JMAP API. The DAV
endpoints advertise `WWW-Authenticate: Basic realm=..., Bearer` but reject
Fastmail OAuth tokens with 401 on every path tried, including the principal
URL — matching Fastmail's documentation, which says an API token is "for JMAP
access" and an app password is "for everything else". `tests/serviceRegistry.test.ts`
asserts the two services never claim the same URL; I confirmed it fails when
they do before trusting it.

### The two supporting commits

Both stand alone and are split out.

`oauth: report why an authorization or token exchange failed` — the callback
handler and the token exchange each discarded the server's own
`error` / `error_description` (RFC 6749 §4.1.2.1 and §5.2) and reported only
that a field was absent. Building this service was close to undebuggable
without it: an `invalid_scope` redirect and an `invalid_target` redirect were
indistinguishable from each other and from a user closing the window. Two of
the three bugs above were invisible until this landed. It also adds an optional
parameter bag to the token and refresh requests, which is how `resource` gets
sent.

`services info: render pattern base API URLs instead of {}` — `baseApiUrls` may
hold a RegExp, and `JSON.stringify` turns one into `{}`. `latchkey services info
aws` has been reporting `"baseApiUrls": [{}]`, i.e. nothing about which hosts it
covers, for precisely the services whose coverage is least obvious. Pre-existing;
it just becomes total for a service whose URLs are all patterns.

### Verification

`fastmail` is exercised end to end against a live account: browser login,
credential check, `getAccount`, and a full mirror of a real mailbox through a
downstream JMAP client — 9 mailboxes, 22 emails, 17 threads, 22 blob downloads,
0 errors. One run touches all three live host shapes: discovery on
`api.fastmail.com`, method calls on `phl.api.fastmail.com`, blobs on
`phl-www.fastmailusercontent.com`. Scope changes are observable in the session
document's `accountCapabilities`, which gained `urn:ietf:params:jmap:contacts`
and `…:calendars` exactly when those scopes were added.

`fastmail-dav` is exercised against a live account too: `checkApiCredentials`
returns 207 Multi-Status, `getAccount` resolves the principal to the account's
e-mail, and an address-book collection enumerates over the stored credential.

Worth knowing, because it cost a long detour: on an account with **no active
subscription** (a lapsed trial), app passwords appeared to be refused for every
legacy protocol while JMAP over an API token or OAuth kept working — and the
refusal is a 401 reading "Incorrect username, password or access token",
indistinguishable from a mistyped password. Subscribing restored DAV with no
change to any code or request. That is one observation rather than documented
Fastmail behaviour, so the service's `info` states it as such and names the
diagnostic that matters: try an IMAP login with the same credential; refused
there too points at the account rather than the credential.

### Notes

- One thing deliberately left alone: if login throws before the browser closes
  (a network error during client registration, say), the `finally` block detaches
  the page/context `close` handlers that would abort the OAuth callback server,
  so its listening socket survives until the 120s timeout rather than being torn
  down immediately. It is self-healing and low impact, and the structure is
  copied verbatim from `services/notion-mcp.ts` — worth fixing in both or
  neither, so I did not fix it here.
- #127 is still worth doing independently: self-registered services can only
  ever carry one literal prefix, which is the general form of the problem this
  service solves for one provider by being built in.
- One rough edge left in `fastmail`: login reuses the `clientId` stored with
  existing credentials, and Fastmail fixes a client's scopes at registration, so
  changing the requested scopes needs `latchkey auth clear fastmail --all`
  first. Documented in the service's `info`. Happy to make it detect and
  re-register instead if you would prefer.


src/services/fastmail-dav.ts
```diff
@@ -0,0 +1,119 @@
...
+   * defensive, so a future shard would not need a new registration.
+   */
+  readonly baseApiUrls = [
+    /^https:\/\/([a-z0-9-]+[.-])?carddav\.fastmail\.com\//,
+    /^https:\/\/([a-z0-9-]+[.-])?caldav\.fastmail\.com\//,
+  ] as const;
+
+  readonly loginUrl = 'https://app.fastmail.com/settings/security';
+
+  readonly info =
```
hynek-urban, line 86: Could you please make this more compact? Ideally hand-written :) For most services, the info string is just a sentence or two. In the ideal case, it's just a pointer to agent-readable documentation, nothing else. Checking the other service implementations should make the convention obvious.

src/services/fastmail.ts
```diff
@@ -0,0 +1,425 @@
...
+   * reachable over JMAP on the hosts below instead.
+   */
+  readonly baseApiUrls = [
+    /^https:\/\/([a-z0-9-]+\.)?api\.fastmail\.com\//,
+    /^https:\/\/([a-z0-9-]+-)?www\.fastmailusercontent\.com\//,
+  ] as const;
+
+  readonly loginUrl = AUTHORIZATION_ENDPOINT;
+
+  readonly info =
```
hynek-urban, line 348: Dtto the previous comment :)

hynek-urban (commented):
Thank you, @thad-imbue ! Before we merge this, would you be willing to also contribute Fastmail permission definitions to Detent? https://github.com/imbue-ai/detent/tree/main/src/schemas/builtin

(We're currently releasing these two in lockstep.)

thad-imbue:
Thanks @hynek-urban! Both addressed.

**Detent permissions** — [imbue-ai/detent#30](https://github.com/imbue-ai/detent/pull/30). Two files, `fastmail.json` and `fastmail-dav.json`, matching the service split here.

One thing there worth your eye: JMAP puts the operation in the request *body*, not the method or path — every call is `POST /jmap/api/` with a `methodCalls` array — so these match on `parsedBody`, following what `notion-mcp.json` already does. And because JMAP batches many calls into one request, the read schemas constrain *every* element of that array; otherwise `["Email/get", "Email/set"]` would slip through a read-only grant. There is a test for that case specifically.

**`info` strings** — cut to the house style in efca712. Each is now a docs link plus the one thing a caller can't guess. Nothing was deleted, only relocated into code comments next to what it explains: the cross-host redirect that eats the Authorization header now sits on `SESSION_ENDPOINT`, the read/write reality of the OAuth scopes on `OAUTH_SCOPES`, and so on. Happy to trim further if they're still long.

thadd3us:
> Thanks @hynek-urban! Both addressed.
> 
> **Detent permissions** — [imbue-ai/detent#30](https://github.com/imbue-ai/detent/pull/30). Two files, `fastmail.json` and `fastmail-dav.json`, matching the service split here.
> 
> One thing there worth your eye: JMAP puts the operation in the request _body_, not the method or path — every call is `POST /jmap/api/` with a `methodCalls` array — so these match on `parsedBody`, following what `notion-mcp.json` already does. And because JMAP batches many calls into one request, the read schemas constrain _every_ element of that array; otherwise `["Email/get", "Email/set"]` would slip through a read-only grant. There is a test for that case specifically.
> 
> **`info` strings** — cut to the house style in [efca712](https://github.com/imbue-ai/latchkey/commit/efca7120cf5e83d82b7047190640b850d86d4f0a). Each is now a docs link plus the one thing a caller can't guess. Nothing was deleted, only relocated into code comments next to what it explains: the cross-host redirect that eats the Authorization header now sits on `SESSION_ENDPOINT`, the read/write reality of the OAuth scopes on `OAUTH_SCOPES`, and so on. Happy to trim further if they're still long.

@hynek-urban Please ignore this, I didn't write it.  I'm going to make a cleanup pass on this PR myself before handing it back to you.

hynek-urban:
@thad-imbue Haha :) Alright, thanks, let me know! I'll ignore all your comments unless they're free from em dashes :)

hynek-urban:
(Although I shouldn't have written this down - the rule probably immediately lost its power :) )

src/services/fastmail-dav.ts
thad-imbue: Much shorter now! :)

PR #132, opened 2026-08-27
Add claude-ai and chatgpt as built-in services with browser login

thad-imbue:
Both were user-registered services until now: datalib and minds each told users
to run `services register` by hand and paste a secret out of DevTools into
`auth set`. minds carried claude.ai as a custom registration of its own
(`additional_services.json`), which only its own gateways knew about and which
needed a propagation mechanism to reach remote ones; chatgpt had no definition
anywhere.

## claude-ai is pure configuration

The web app authenticates with a single `sessionKey` cookie, so this configures
the existing generic `cookie-capture` flow rather than implementing anything.
The credential it yields is byte-identical to what
`auth set -H "Cookie: sessionKey=..."` stores today, so existing accounts keep
working — and `loadRegisteredServicesIntoServiceRegistry` already skips a
user-registered name that collides with a built-in, so anyone who registered
`claude-ai` by hand transparently upgrades rather than needing a migration.

## chatgpt cannot use that flow

Its session cookie authenticates the *page*, which then calls
`/api/auth/session` to mint the short-lived bearer token that `/backend-api`
wants. Capturing the cookie would store something the API refuses. So the
session watches for that response and lifts `accessToken` out of the body —
the same value a user reads out of DevTools today, obtained without DevTools.
A signed-out visitor gets `{}` from the same endpoint, so the presence of the
token is the completion signal, not the request having happened.

There is no refresh path: the refresh material is the cookie, which stays in
the browser profile and never reaches us. `credentialStatus` going `invalid`
is routine rather than a fault; signing in again mints a fresh token.

## Effect

Both now report `authOptions ["browser", "set"]`. With this branch, five of the
seven services datalib mirrors from have a browser login (slack, github,
fastmail, claude-ai, chatgpt); notion and gitlab remain `set`-only.

## Verified against the live services

With `LATCHKEY_CURL` pointed at a Chrome-impersonating curl:

- `/api/organizations` and `/backend-api/me` answer 200 and drive the
  credential check.
- `/api/account` carries `email_address` and `/backend-api/me` carries `email`,
  which is what `getAccount` reads — so accounts get named instead of landing
  on the unnamed default. (`/api/auth/current_account` does *not* exist; it
  404s.)
- `services info` reports `authOptions ["browser", "set"]` and `valid` for both,
  against credentials stored before this branch existed.

**Not verified end to end:** the interactive browser login itself, and
therefore `getAccount` in situ — the account is only derived at login time and
there is no re-derive command, so both need a real sign-in at a keyboard.

## Cloudflare caveat, worth a maintainer's opinion

Both hosts sit behind Cloudflare's managed challenge, which answers vanilla
curl with a 403 interstitial. The browser login is unaffected — it drives a
real Chrome — but `checkApiCredentials` and `getAccount` go out over curl, so
with a stock curl they report `invalid` and no account for credentials that are
fine. Everything works once `LATCHKEY_CURL` points at an impersonating curl.

This is documented on both services rather than worked around. Adding
`X-Imbue-Impersonate: 1` to their `credentialCheckCurlArguments` would make the
check correct wherever the dispatch curl is present — which is both datalib and
minds — at the cost of putting a consumer-specific header in latchkey. That
felt like a maintainer's call rather than something to slip into this PR.

## Testing

`npm run lint` clean; `npx vitest run` 790 passed / 3 skipped, including 9 new
tests that drive both sessions through the public login path with responses fed
in by hand.


hynek-urban:
@thad-imbue Thank you! I think this PR is unfortunately out of scope for Latchkey at the moment because it targets undocumented, private APIs. I think we should create a plugin system for Latchkey so that we can make use of contributions like this for people who really want that without making them part of the core.

PR #133, opened 2026-08-27
Reload registered services on every gateway request

qi-imbue:
## What changes for users

A running `latchkey gateway` built its service registry once, at startup, and never looked at `config.json` again. So `latchkey services register` on the gateway host wrote the entry to disk, but the gateway kept rejecting that service until someone restarted it.

After this, registering and deregistering a service take effect on the next request, with no restart. That matches how the gateway already treats the rest of its on-disk state:

| State | Reloaded per request before this PR? |
|---|---|
| Stored credentials (`auth set`, refreshed tokens) | yes |
| `permissions.json` | yes |
| Which permissions file a request uses | yes |
| Registered custom services | **no** — this PR |

Nothing else changes. The `settings` section of `config.json`, `hideBuiltinServices` included, is still read once at startup and still needs a restart. #134 is the follow-up that reloads the whole configuration, if we'd rather not reload only part of it.

## Cost

Measured end to end, not just the file read: a real gateway via `startGateway` with a stubbed curl runner, 2000 requests after a 300-request warmup, three runs of this branch against three of a variant that builds the registry once. Median of the p50s, on an M-series Mac:

| registered services in config.json | registry built once | rebuilt per request | added |
|---|---|---|---|
| 0 | 337 µs | 361 µs | **+24 µs** |
| 10 | 322 µs | 358 µs | **+36 µs** |
| 100 | 325 µs | 455 µs | **+130 µs** |

About 25 µs fixed, plus ~1.3 µs per registered service. Total gateway overhead per request is ~330 µs here, and that excludes the `curl` spawn a real proxied request pays (~9.7 ms before any network) — so this is well under 1% of a real request's wall time.

Each request gets its own registry rather than sharing a rebuilt one, since a shared one would change underneath requests already in flight.

## Internal changes worth knowing about

**`SERVICE_REGISTRY` is gone from the package exports.** Building a registry is one function, `createServiceRegistry`, used by both the CLI at startup and the gateway per request, rather than the same load-then-hide pair written out in each place. The CLI therefore builds its own registry instead of mutating the `SERVICE_REGISTRY` singleton, which left that singleton with a single consumer — so it now lives in `src/migrations.ts` as an unexported constant and is no longer re-exported from `src/index.ts`.

That breaks the programmatic API for anyone importing it; per review, only CLI compatibility is maintained. Migrations are unaffected in substance: they ran before the old registry load, so they already only ever saw built-in services.

This also let `src/cli.ts` drop its registry setup entirely.

## Manual testing

Run `latchkey gateway` in one terminal and work in another. None of items 1–4 should need a restart.

1. **Register while running** — `latchkey services register my-thing --base-api-url https://…/`, give it credentials with `latchkey auth set`, then make a request through `/gateway/` and confirm the credentials are injected. Previously this returned 400 until the gateway was restarted.
2. **Deregister while running** — `latchkey services deregister my-thing`, then repeat the request and confirm it is no longer served.
3. **Self-hosted instance** — register one with `--service-family` (e.g. a self-hosted GitLab) and confirm the family's credential handling applies to it through the gateway.
4. **Built-ins unaffected** — with a registered service present, confirm a built-in service still resolves and injects normally.
5. **Settings still need a restart** — add `"hideBuiltinServices": ["github"]` to `config.json` while the gateway runs and confirm it does *not* take effect until restart. This is the deliberate scope boundary, and the behavior #134 changes.

The rest exercise the startup restructure, which touches every CLI invocation and not just the gateway:

6. **Plain CLI, no gateway** — `latchkey services list` shows registered services, `latchkey curl` against one injects credentials, and `latchkey services list --builtin` excludes them. Confirm `hideBuiltinServices` in `config.json` still hides a service at startup.
7. **CLI pointed at a remote gateway** — with `LATCHKEY_GATEWAY` set, confirm `latchkey services list` still reports the services the *gateway* knows about, and that `latchkey services register` is still refused locally. This is the path where the CLI deliberately skips `config.json`, since the gateway owns registered services there.
8. **Migrations on an old data directory** — the registry is now built before migrations run rather than after. Migrations read the untouched `SERVICE_REGISTRY` in both cases, so this should be unchanged, but it is worth confirming against a pre-migration `~/.latchkey`.
9. **Under load** — hammer the gateway while registering and deregistering a service, and confirm no crashes and no torn behavior.

## Automated tests

- `tests/serviceRegistry.test.ts` — 8 cases for `createServiceRegistry`: built-ins plus config, missing config file, a service added between calls, one removed between calls, registries independent of each other, a null config path skipping the file while still hiding, hiding, and a registered service whose family is hidden.
- `tests/gateway.test.ts` — 4 integration cases against a live gateway, covering items 1, 2 and 4 above, plus the case where the deregistered service is still sitting in the registry the gateway was handed.

Verified the three behavioral tests fail against the unmodified request path. Full suite: 821 passed, lint and prettier clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,14 +365,6 @@ route your agent's `latchkey` calls to the host Latchkey
 agent won't be able to tamper with your Latchkey configuration
 while still being able to use Latchkey itself as intended.
 
-A running gateway reads the registered services from
-`config.json` on every request, so `latchkey services register`
-and `latchkey services deregister` take effect immediately, the
-same way stored credentials and `permissions.json` do. The
-`settings` section of `config.json` is still read once at
-startup, so changing a setting (including `hideBuiltinServices`)
-requires restarting the gateway.
-
 #### Password
 
 You can additionally protect the gateway with a shared password.
--- a/README.md
+++ b/README.md
@@ -365,6 +365,14 @@ route your agent's `latchkey` calls to the host Latchkey
 agent won't be able to tamper with your Latchkey configuration
 while still being able to use Latchkey itself as intended.
 
+A running gateway reads the registered services from
+`config.json` on every request, so `latchkey services register`
+and `latchkey services deregister` take effect immediately, the
+same way stored credentials and `permissions.json` do. The
+`settings` section of `config.json` is still read once at
+startup, so changing a setting (including `hideBuiltinServices`)
+requires restarting the gateway.
+
 #### Password
 
 You can additionally protect the gateway with a shared password.
--- a/src/gateway/server.ts
+++ b/src/gateway/server.ts
@@ -18,6 +18,7 @@ import {
 } from './gatewayEndpoint.js';
 import { handleLatchkeyRequest } from './latchkeyEndpoint.js';
 import { GATEWAY_PASSWORD_HEADER, passwordsMatch } from './password.js';
+import { createServiceRegistry } from '../serviceRegistry.js';
 import {
   dispatchExtensionRequest,
   loadExtensions,
@@ -155,12 +156,27 @@ export async function startGateway(
       return;
     }
 
+    // Build the registry from config.json as it reads now, rather than using
+    // the one in `deps`, so that registering or deregistering a service while
+    // the gateway runs takes effect without a restart — the way writes to the
+    // credential store and to permissions.json already do. Each request gets
+    // its own: a shared one would be rebuilt underneath requests already in
+    // flight.
+    const requestDeps: CliDependencies = {
+      ...deps,
+      registry: createServiceRegistry(
+        deps.builtinServices,
+        deps.config.configPath,
+        deps.config.hideBuiltinServices
+      ),
+    };
+
     // Latchkey RPC endpoint
     if (rawUrl === '/latchkey/' || rawUrl === '/latchkey') {
       const requestPromise = handleLatchkeyRequest(
         request,
         response,
-        deps,
+        requestDeps,
         apiCredentialStore,
         encryptedStorage
       ).catch((error: unknown) => {
@@ -192,7 +208,7 @@ export async function startGateway(
         request,
         response,
         targetUrl,
-        deps,
+        requestDeps,
         apiCredentialStore,
         options
       ).catch((error: unknown) => {
@@ -212,7 +228,7 @@ export async function startGateway(
     }
 
     // Finally, try extensions (if any).
-    const requestPromise = runExtensions(request, response, extensions, deps, options)
+    const requestPromise = runExtensions(request, response, extensions, requestDeps, options)
       .then((handled) => {
         if (!handled && !response.headersSent) {
           response.writeHead(404);
--- a/src/serviceRegistry.ts
+++ b/src/serviceRegistry.ts
@@ -275,11 +275,3 @@ export const BUILTIN_SERVICES: readonly Service[] = [
   OPENROUTER,
   TAILSCALE,
 ];
-
-/**
- * The built-in services alone, with nothing from config.json and nothing
- * hidden. The CLI and the gateway both build their own registry with
- * {@link createServiceRegistry} instead; this is for callers that want the
- * services latchkey ships with.
- */
-export const SERVICE_REGISTRY = new ServiceRegistry(BUILTIN_SERVICES);
--- a/src/serviceRegistry.ts
+++ b/src/serviceRegistry.ts
@@ -212,7 +212,32 @@ export function loadRegisteredServicesIntoServiceRegistry(
   }
 }
 
-export const SERVICE_REGISTRY = new ServiceRegistry([
+/**
+ * Build a registry holding `builtinServices` plus the services registered in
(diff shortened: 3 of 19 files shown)


src/serviceRegistry.ts
```diff
@@ -249,4 +274,12 @@ export const SERVICE_REGISTRY = new ServiceRegistry([
...
-]);
+];
+
+/**
+ * The built-in services alone, with nothing from config.json and nothing
+ * hidden. The CLI and the gateway both build their own registry with
+ * {@link createServiceRegistry} instead; this is for callers that want the
+ * services latchkey ships with.
+ */
+export const SERVICE_REGISTRY = new ServiceRegistry(BUILTIN_SERVICES);
```
qi-imbue: This is now only used by migration.ts, so we can technically make it an unexported symbol there.

But removing an exported symbol technically breaks any external code that consumes Latchkey's API. Is this something we care about?

qi-imbue (commented):
I didn't read the tests, but the implementation code looks quite reasonable.

qi-imbue:
If we think it's a bit weird to only reload part of `config.ts` and prefer reloading it entirely, I also made https://github.com/imbue-ai/latchkey/pull/134. (I haven't reviewed it myself though.)

README.md
```diff
@@ -365,6 +365,14 @@ route your agent's `latchkey` calls to the host Latchkey
 agent won't be able to tamper with your Latchkey configuration
 while still being able to use Latchkey itself as intended.
 
+A running gateway reads the registered services from
```
hynek-urban: Let's not add this to the README for now. (There are many undocumented aspects of the gateway already - we really should create better documentation soon but until we have that, it's better for the README to stay lean - I mean, it's already too long :( )

src/serviceRegistry.ts
hynek-urban: No, I'm only considering the CLI backwards compatibility. I don't expect Latchkey to be used as a library and refuse to maintain the programmatic API surface :)

src/gateway/server.ts
```diff
@@ -155,12 +156,27 @@ export async function startGateway(
       return;
     }
 
+    // Build the registry from config.json as it reads now, rather than using
+    // the one in `deps`, so that registering or deregistering a service while
+    // the gateway runs takes effect without a restart — the way writes to the
+    // credential store and to permissions.json already do. Each request gets
+    // its own: a shared one would be rebuilt underneath requests already in
+    // flight.
+    const requestDeps: CliDependencies = {
```
hynek-urban, line 165: Before we merge this, can we measure how much overhead this adds to each proxied request?

hynek-urban (commented):
@qi-imbue Thanks! I changed the base to `dev` (because I'll publish it together with some other changes) and left you a few comments below. Feel free to merge as soon as you're happy with it.

README.md
qi-imbue: Done — reverted in fd48bbe, the README is untouched by this branch now (and by #134).

---
*Posted by Claude on behalf of @qi-imbue.*

src/serviceRegistry.ts
qi-imbue: Done in fd48bbe. `SERVICE_REGISTRY` moved into `migrations.ts` as an unexported constant and is out of the package exports.

The tests and `scripts/recordBrowserSession.ts` also used it, so they build their own registry over `BUILTIN_SERVICES` — the six tests that want one share `tests/builtinServiceRegistry.ts`.

---
*Posted by Claude on behalf of @qi-imbue.*

src/gateway/server.ts
qi-imbue: Measured. I started a real gateway via `startGateway` with a stubbed curl runner (so this is the gateway's own work, with no subprocess or network in the way), fired 2000 requests through `/gateway/` after a 300-request warmup, and ran the whole thing three times against this branch and three times against a variant that builds the registry once at startup. Median of the p50s, on an M-series Mac:

| registered services in config.json | registry built once | rebuilt per request | added |
|---|---|---|---|
| 0 | 337 µs | 361 µs | **+24 µs** |
| 10 | 322 µs | 358 µs | **+36 µs** |
| 100 | 325 µs | 455 µs | **+130 µs** |

So it is roughly 25 µs of fixed cost (one `readFileSync` plus building a registry over the built-in services) and about 1.3 µs per registered service on top.

For context: total gateway overhead per request is ~330 µs here, and that excludes the `curl` spawn a real proxied request pays, which I measured separately at ~9.7 ms before any network. So this adds ~7% to the gateway's own work at a typical config size, and well under 1% of the wall time of an actual proxied request.

If the 100-service number bothers you, the obvious cheap fix is an mtime check before rebuilding (`statSync` is a few µs), which would make the steady-state cost roughly zero and only pay the rebuild when the file actually changes. I left it out to keep the change simple, but it is a small addition if you would prefer it.

---
*Posted by Claude on behalf of @qi-imbue.*

hynek-urban: No, I think this is fine, thanks :) That's much lower than I thought.

PR #134, opened 2026-08-27
Reload the whole configuration on every gateway request

qi-imbue:
Stacked on #133, so this diff is against that branch. #133 alone reloads only the registered services; this one reloads the whole configuration, which is the "if we'd rather not reload just part of it" option.

## What changes for users

A running `latchkey gateway` rereads the environment and `config.json` on every request, instead of holding whatever they said when it started. Editing a setting now takes effect on the next request, with no restart.

These become live:

| Setting | Effect while running |
|---|---|
| `passthroughUnknown` | requests to unknown services start or stop being forwarded |
| `hideBuiltinServices` | adding a name hides that service; removing it brings the service back |
| `permissionsConfig` | the gateway starts consulting a different `permissions.json` |
| `permissionsDoNotUseBuiltinSchemas` | built-in request schemas stop or start being applied |
| `credentialsRefreshDisabled` | expired credentials stop or start being refreshed |
| `curlCommand` | a different curl binary is spawned, for proxied requests and credential checks alike |
| `browserDisabled`, `browserEphemeral`, `appNamePrefix` | apply to `auth browser` issued through the gateway |

Stored credentials, `permissions.json` and (from #133) registered services already reloaded this way.

## What still needs a restart

These control work that finishes before the first request arrives, so a running server genuinely cannot apply them:

| Setting | Why |
|---|---|
| `keyringServiceName`, `keyringAccountName` | the encryption key was resolved at startup |
| `gatewayListenHost`, `gatewayListenPort` | the listening socket is already bound |
| `countingDisabled` | the daily count already ran |

Rather than let these change silently, the gateway writes a warning to stderr the first time it notices one differ, naming the reason and saying a restart is needed. It warns once per setting, not once per request.

## Internal changes worth knowing about

**`curl.ts` gained process-wide state.** `curlCommand` is read at 8 call sites deep inside the services — `src/services/core/base.ts:224` sits inside `Service.checkApiCredentials`, a public method many services override — so threading a config down to them would change that signature and every override. Instead the gateway hands the command to `src/curl.ts`, next to the swappable runners that module already keeps. This is the one piece I'd want a second opinion on; dropping it and leaving `curlCommand` restart-only removes about 30 lines.

**`Config.reload()`** returns a new `Config` from the same sources the receiver was built from, leaving the receiver alone so a request in flight keeps a stable view. `Config` gained two public readonly fields (`getEnv`, `loadSettingsFromFile`) to make that possible.

## Risk to be aware of

A malformed `config.json` no longer fails loudly at startup — it is now read on every request, and both `readConfigFile` and `loadSettings` fall back to defaults rather than throwing. A half-written file (an editor that truncates before writing) would therefore make the gateway briefly serve *default* settings, silently. Latchkey's own writes go through `writeFileAtomic`, so `latchkey services register` and friends are safe; hand-editing is the exposure.

## Manual testing

Run `latchkey gateway` in one terminal and watch its output while doing the following in another. None of these should need a restart.

1. **passthroughUnknown** — `curl` an unknown host through `/gateway/`; expect 400. Set `"passthroughUnknown": true` under `settings` in `config.json`, repeat; expect the request to be forwarded. Set it back to `false` and confirm it 400s again.
2. **hideBuiltinServices** — with credentials stored for a built-in service, make a request through the gateway and confirm it works. Add `"hideBuiltinServices": ["github"]`, repeat; expect 400. Remove the name and confirm the service comes back.
3. **permissionsConfig** — point it at a second `permissions.json` that denies something the first allows, and confirm the denial takes effect on the next request.
4. **curlCommand** — point it at a wrapper script that logs its arguments and then execs the real curl. Confirm the log records both a proxied `/gateway/` request **and** a `latchkey services info <service>` issued through the gateway (that second one is the deep call path, and is the reason for the `curl.ts` change).
5. **Restart-only warning** — change `gatewayListenPort` in `config.json`, then make two requests. Expect exactly one warning on the gateway's stderr naming `gatewayListenPort`, and the gateway still listening on the original port.
6. **No spurious warnings** — change only `passthroughUnknown` and confirm nothing is warned about.
7. **Registered services still work** (carried over from #133) — `latchkey services register` a self-hosted instance while the gateway runs, use it immediately, then `latchkey services deregister` it and confirm it stops being served.
8. **Malformed config** — write invalid JSON into `config.json` while the gateway is running. Expect it to keep serving on default settings rather than crash, and to recover once the file is valid again.
9. **Under load** — hammer the gateway (`for i in $(seq 200); do ... done` or similar) while editing `config.json`, and confirm no crashes, no torn behavior, and that requests keep succeeding.

## Automated tests

- `tests/config.test.ts` — 5 cases for `reload`: setting added, setting removed, receiver left untouched, environment still winning over the file, repeated reloads.
- `tests/gateway.test.ts` — 7 integration cases against a live gateway, covering items 1, 2, 4, 5 and 6 above.
- `tests/curl.test.ts` — new file, 3 cases for the curl-command seam.

Verified that 6 of the 7 gateway cases fail when the `reload()` call is reverted. Full suite: 836 passed, lint and prettier clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,13 +365,22 @@ route your agent's `latchkey` calls to the host Latchkey
 agent won't be able to tamper with your Latchkey configuration
 while still being able to use Latchkey itself as intended.
 
-A running gateway reads the registered services from
-`config.json` on every request, so `latchkey services register`
-and `latchkey services deregister` take effect immediately, the
-same way stored credentials and `permissions.json` do. The
-`settings` section of `config.json` is still read once at
-startup, so changing a setting (including `hideBuiltinServices`)
-requires restarting the gateway.
+A running gateway reads the environment and `config.json` again on
+every request, so most changes take effect without a restart:
+services added or removed with `latchkey services register` and
+`latchkey services deregister`, and settings such as
+`passthroughUnknown`, `hideBuiltinServices`, `permissionsConfig`,
+`credentialsRefreshDisabled`, `browserDisabled` and `curlCommand`.
+Stored credentials and `permissions.json` were already read this
+way.
+
+The exceptions are the settings whose work is done before the
+first request arrives: `keyringServiceName` and
+`keyringAccountName` (the encryption key has already been
+resolved), `gatewayListenHost` and `gatewayListenPort` (the socket
+is already bound), and `countingDisabled` (the daily count has
+already run). The gateway logs a warning when it sees one of these
+change, and applying it needs a restart.
 
 #### Password
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -185,6 +185,13 @@ function resolveBoolean(envValue: string | undefined, fileValue: boolean | undef
  * LATCHKEY_DIRECTORY and LATCHKEY_ENCRYPTION_KEY are env-only.
  */
 export class Config {
+  /**
+   * Where this config was read from, kept so that {@link reload} can consult
+   * the same sources again.
+   */
+  readonly getEnv: (name: string) => string | undefined;
+  readonly loadSettingsFromFile: (configPath: string) => Settings;
+
   readonly directory: string;
   readonly curlCommand: string;
   /**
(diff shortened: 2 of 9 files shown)


hynek-urban:
Thanks, @qi-imbue . I don't feel that great about this one - I feel that it makes things slightly more fragile and harder to reason about. Almost all services / daemons I can think of expect you to restart them explicitly when the base configuration changes.

(The case of registered services, just like credentials, is a bit different - those rather correspond to "data" the program operates on than fundamental configuration. At least that's how I perceive it.)

PR #139, opened 2026-08-28
Classify unreadable stored credentials, and stop them failing unrelated exports

pseay-imbue:
## Why

In production, an app that bundles Latchkey 3.5.0 shares a credential store with newer standalone CLIs. One of those wrote a `tailscale`-typed credential, whose `objectType` the bundled binary's schema does not know. From then on **every** `latchkey auth re-encrypt` from that store failed -- including exports that only asked for `--services slack` -- because `re-encrypt` called `ApiCredentialStore.getAll()` (which schema-validates all stored credentials) *before* applying the `--services` filter.

The caller could only tell this failure from any other by matching the message text (`Invalid credential data for service <name>:`), since every `re-encrypt` failure path exits 1 through the same `deps.exit(1)` with no structured output.

## What changed

1. **`re-encrypt` parses only what it exports.** Stored service names now come from `ApiCredentialStore.listServiceNames()` (raw keys, no parsing) plus `listPreparedServiceNames()`; the credentials and preparations are parsed via the new `getAllForServices()` once the selection is known, before anything is written. A credential type this binary does not understand no longer breaks exports of other services.

2. **The remaining failure is classified.** `parseCredentials` now throws `InvalidStoredCredentialError` (an `ApiCredentialStoreError` subclass carrying `serviceName`). When a *requested* service is unreadable, `re-encrypt` reports it through the new `src/cliErrors.ts`:
   - exit code **3** (`INVALID_STORED_CREDENTIAL_EXIT_CODE`), alongside the existing `PERMISSION_DENIED_EXIT_CODE` precedent;
   - with `LATCHKEY_ERROR_JSON=1`, a JSON object on stderr after the human message: `{"latchkeyError": {"code": "invalid_stored_credential", "message": "...", "serviceName": "tailscale"}}`.

   Human-readable output is unchanged, and the JSON line only appears when the env var opts in, so nothing existing sees different output.

3. README documents `LATCHKEY_ERROR_JSON` and the dedicated exit code.

## Tests

`tests/cli.test.ts`, in the `auth re-encrypt` block: an unreadable credential no longer fails an export of another service; a requested unreadable credential exits 3; and with `LATCHKEY_ERROR_JSON=1` the JSON object carries the code and service name.

`npm run lint`, `npm run typecheck`, `npm test`: 859 passed, 7 skipped, 1 todo, 0 failed.

## Consumer

imbue-ai/mngr-internal#731 currently classifies this failure with a regex over stderr. Once this ships in a release, that PR drops the regex and dispatches on the exit code plus the JSON payload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,7 @@ defaults:
 - `LATCHKEY_PERMISSIONS_CONFIG`: override the `permissions.json` location.
 - `LATCHKEY_PERMISSIONS_DO_NOT_USE_BUILTIN_SCHEMAS`: do not use the built-in permission definitions.
 - `LATCHKEY_PASSTHROUGH_UNKNOWN`: if set, Latchkey will forward requests (via `latchkey curl` or gateway) even if no credentials are injected.
+- `LATCHKEY_ERROR_JSON`: when set to `1`, a failure that Latchkey can classify is also printed on stderr as a JSON object (`{"latchkeyError": {"code": ..., "message": ..., "serviceName": ...}}`) after the human-readable message. Programs driving Latchkey as a subprocess can dispatch on `code` instead of matching on message text. Classified failures also use a dedicated exit code (currently: `3` for a stored credential no known schema accepts, which usually means a newer Latchkey wrote the store).
 - `LATCHKEY_HIDE_BUILTIN_SERVICES`: comma-separated list of built-in service names to disable/hide.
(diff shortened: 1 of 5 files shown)


hynek-urban:
Thanks @pseay-imbue . I'm going to close this for now, for several reasons: the re-encrypt logic has been updated in the meantime; I think the tool should bail if it encounters illegible data from a future version; and I don't think we should extend the surface with the new `LATCHKEY_ERROR_JSON` env var.

There are some good ideas here, though, so I'll probably follow-up with something inspired by this!

PR #143, opened 2026-09-03
token-capture: ask the mint endpoint, don't only wait to overhear it

thad-imbue:
#140 shipped `token-capture` with a gap I should have caught, and I'm sorry: the flow only ever *watched* for the mint endpoint. That works for an app that calls its own endpoint while we happen to be looking, and it silently never completes for one that doesn't — which turns out to include the single most obvious service anyone would point this flow at. The PR was written and tested entirely against hand-fed responses, so nothing in it ever exercised the assumption it rested on. This adds the missing half.

## The failure

Registering chatgpt.com exactly as #140's own `--help` example suggests produces a login that hangs forever:

```bash
latchkey services register chatgpt-tc \
  --base-api-url="https://chatgpt.com/" \
  --login-url="https://chatgpt.com/auth/login" \
  --login-flow=token-capture \
  --login-flow-params='{"tokenUrl": "https://chatgpt.com/api/auth/session",
                        "tokenField": "accessToken"}'
```

Sign-in completes, the app loads, nothing happens. Instrumenting the browser showed why: a full bootstrap of the signed-in app makes **48 requests and not one of them is to `/api/auth/session`**. The endpoint is not gone — asking it directly returns a 1984-character `accessToken` — but a web app calls its mint endpoint when *it* wants a token, not when we want one. NextAuth conventions say where the endpoint is, not that the app will ever call it.

It also never fails. Every unmet condition returns null and the wait loop has no timeout, so "endpoint moved", "body unreadable", "field absent" and "never called" all present as one indistinguishable hang. That is a bad property on its own and it is what made this take a while to pin down.

## The fix

The flow now also asks. Once the browser reaches the endpoint's origin, it is requested from inside the page every couple of seconds. Watching stays — it costs nothing and is the whole story for an app that does call its endpoint — and whichever path produces a token first ends the login.

`ServiceSession` gains `whileWaitingForLogin`, the active counterpart to the existing `onResponse`: work a session does itself between polls of `isLoginComplete`. The default does nothing, so no other flow changes.

## Three choices worth the review

- **`page.evaluate(fetch)` rather than Playwright's `context.request`.** The latter shares the cookie jar but not the browser's TLS fingerprint, which is exactly what Cloudflare inspects — and this flow's whole audience is apps sitting behind that kind of wall. Making the page issue the request means it carries the same session *and* the same clearance the app itself has.
- **Waiting for the page to reach the endpoint's origin** is not just tidiness. The response read would be cross-origin otherwise and refused, and mid-login the browser is usually parked on an identity provider where there is nothing to ask.
- **`whileWaitingForLogin` implementations swallow their own transient failures.** Anything thrown there aborts the login, and a page navigating out from under a request is routine during a sign-in, not fatal. The contract is stated on the hook.

Also separated: `response.text()` throwing on a body the page had navigated past was being caught by the same `catch` as a JSON parse failure. Two different situations, now two branches.

## Testing

`npm run lint` and `npx prettier --check` clean.

7 new tests in `tests/tokenCapture.test.ts`, driving the hook through the public path with a page whose replies are canned, mirroring how the existing tests drive responses. **4 of them fail against the passive-only flow** — I neutered the override and watched them go red before believing them, because a test for a silent hang that cannot itself fail is worth nothing.

`npx vitest run` → 876 passed, 8 failed. The same 8 fail on an unmodified `origin/dev` in this environment (subprocess tests in `cli.test.ts`, `devShim.test.ts`, `lint.test.ts` and `typecheck.test.ts` hitting a 20s timeout), so they are not from this change. #140 reported 6 of these; `lint`/`typecheck` have since joined, same cause.

What the unit tests cannot cover is the body of the `page.evaluate` callback, which only runs in a browser. So this was verified live: `auth browser chatgpt-tc` against a local build now completes within about two seconds of the chat UI appearing, and the credential it stores works —

| endpoint | status |
|---|---|
| `/backend-api/me` | 200, real profile payload |
| `/backend-api/conversations?offset=0&limit=3` | 200, 3 items |

(through a JA3-impersonating curl, since plain curl gets a 403 from Cloudflare on chatgpt.com regardless of credentials.)

## Two things noticed, deliberately not fixed here

- `RegisteredService.getAccount()` returns null unconditionally, so every login for a registered service lands in the unnamed default account and overwrites the last one. Registering a service twice is currently the only way to hold two accounts for it.
- The login wait loop has no timeout, which is what turns any of this flow's failure modes into an indefinite hang with nothing on stderr.

Happy to take either as a follow-up if you'd like them.

### DIFF
--- a/src/services/core/base.ts
+++ b/src/services/core/base.ts
@@ -350,11 +350,32 @@ export abstract class ServiceSession {
   ): Promise<ApiCredentials | null>;
 
   /**
+   * Work the session does itself while the login phase is still going, run
+   * between polls of {@link isLoginComplete}.
+   *
+   * The default does nothing, which is right for a session that only has to
+   * react to what the browser does. A session whose credential the page never
+   * hands over unprompted overrides this and goes and gets it.
+   *
+   * Implementations own their transient failures: anything thrown here aborts
+   * the login, so a step that can fail and be retried on the next poll has to
+   * swallow that itself.
+   *
+   * Public for the same reason {@link onResponse} is: it is one half of the
+   * pair of hooks a login phase drives, and the tests drive it directly rather
+   * than through a browser.
+   */
+  whileWaitingForLogin(_page: Page): Promise<void> {
+    return Promise.resolve();
+  }
+
+  /**
    * Wait until the browser login phase is complete.
    */
   private async waitForLoginComplete(page: Page): Promise<void> {
     while (!this.isLoginComplete()) {
       await page.waitForTimeout(100);
+      await this.whileWaitingForLogin(page);
     }
   }
 
--- a/src/services/core/loginFlows/tokenCapture.ts
+++ b/src/services/core/loginFlows/tokenCapture.ts
@@ -1,7 +1,14 @@
 /**
  * A generic browser login for services whose credential is a token the web app
- * mints for itself: open a URL, watch the responses that arrive while the user
- * signs in, and lift a token out of the JSON body of a named endpoint.
+ * mints for itself: open a URL, and lift a token out of the JSON body of a
+ * named endpoint, either as it goes by or by asking for it.
+ *
+ * Watching alone is not enough, because a web app calls its own mint endpoint
+ * when it needs a token and not otherwise. ChatGPT is the case that showed
+ * this: it signs in, renders the app, and never calls `/api/auth/session` —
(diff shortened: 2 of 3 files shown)


hynek-urban:
@thad-imbue LGTM, thanks!

PR #144, opened 2026-09-03
Add a gateway header that forwards a request without credential injection

qi-imbue:
## Summary

Adds `X-Latchkey-Gateway-No-Credentials`, a gateway-internal header. A `/gateway/<url>` request carrying it (any value) is forwarded exactly as received: no service lookup, no credential injection, and no permission check. The gateway password, when configured, is still enforced, and the header is stripped before the request goes upstream, like the other `X-Latchkey-Gateway-*` headers.

## Why

minds is moving to per-machine credential stores: a remote workspace's VPS gateway holds the credentials for that workspace, and the desktop gateway holds only the desktop's. Some destinations block datacenter IPs, so a remote workspace sometimes needs a request to leave from the user's computer. Until now that worked by having the desktop gateway inject credentials from the desktop store; once the stores are split, the desktop no longer has the right credentials.

The new flow has the VPS gateway inject its own credentials and run the permission check as for any other request, then has its curl send the finished request on to the desktop gateway with this header. The desktop gateway only lends its network position. There is nothing of the desktop's store to protect on such a request, which is why the permission check is skipped too (minds' per-account grants are keyed on the injected account, which the desktop no longer knows, so a check there would deny everything).

## Changes

- `src/gateway/gatewayEndpoint.ts`: `GATEWAY_NO_CREDENTIALS_HEADER` joins `GATEWAY_INTERNAL_HEADERS` (so `buildCurlArguments` strips it) and, when present, `handleGatewayRequest` skips `prepareCurlInvocation` and hands the rebuilt curl arguments straight to curl.
- `tests/gateway.test.ts`: the header is stripped from curl arguments; a marked request keeps the client's own `Authorization` and gets no stored one; the permission check is not consulted; an unknown service is fine; the password is still required; an unmarked request is unaffected.

`npm run lint` and `npm test` pass (881 tests).

Companion changes: imbue-ai/datalib (the dispatch curl that sets this header) and the minds monorepo (which will bump the pinned latchkey version once this is released).

https://claude.ai/code/session_015TyyvhggzVVgyLej9tHdfV


src/gateway/gatewayEndpoint.ts
```diff
@@ -343,27 +354,36 @@ export async function handleGatewayRequest(
   const accountHeader = request.headers[GATEWAY_ACCOUNT_HEADER];
   const account = typeof accountHeader === 'string' ? accountHeader : undefined;
 
+  // A request marked as carrying its own credentials bypasses the injection
+  // pipeline entirely (see `GATEWAY_NO_CREDENTIALS_HEADER`): there is nothing
+  // of this gateway's to protect, so neither the service lookup nor the
+  // permission check runs. The marker itself was stripped along with the other
+  // gateway-internal headers above.
+  const isNoCredentialsRequest = request.headers[GATEWAY_NO_CREDENTIALS_HEADER] !== undefined;
```
qi-imbue: Not sure if you have an opinion on whether this should look for some specific value or accept anything.

qi-imbue:
Tested manually:

```sh
# works
env LATCHKEY_GATEWAY=http://localhost:1989 LATCHKEY_GATEWAY_PASSWORD=testpw LATCHKEY_DISABLE_COUNTING=1 npx tsx src/cli.ts curl -s -H "X-Latchkey-Gateway-No-Credentials: 1" -H "Authorization: Bearer x" https://httpbin.org/headers
# fails with "no service matches url"
env LATCHKEY_GATEWAY=http://localhost:1989 LATCHKEY_GATEWAY_PASSWORD=testpw LATCHKEY_DISABLE_COUNTING=1 npx tsx src/cli.ts curl -s -H "Authorization: Bearer x" https://httpbin.org/headers
```

src/gateway/gatewayEndpoint.ts
hynek-urban: No, I think this is fine :)

hynek-urban:
@qi-imbue Thanks! I added two changes:

- Permission checks should be done even for requests where no credentials are added.
- And we should refuse arbitrary credential-less requests unless LATCHKEY_PASTHROUGH_UKNOWN is set (otherwise we'd introduce a backwards incompatible change - in theory, someone could rely on that as a security measure).

PR #158, opened 2026-09-18
Optionally tell curl which service was matched

qi-imbue:
The `LATCHKEY_DIAGNOSTIC_HEADERS` env causes Latchkey to call curl with additional diagnostic headers. This currently just includes `X-Latchkey-Matched-Service`.


hynek-urban:
@qi-imbue Looks good to me, thanks!

